PR #9, opened 2019-02-13
Fix sorting of nodes by timestamp

adambalogh:


internal/extender/resource.go
```diff
@@ -196,24 +196,27 @@ func (s *SparkSchedulerExtender) fitEarlierDrivers(
 	return true
 }
 
-func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v1.Pod, nodeNames []string) (string, string, error) {
+func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v1.Pod, nodeNames set) (string, string, error) {
```
onursatici: I think we should preserve the signature and inline the convertToSet method below

internal/extender/resource.go
```diff
@@ -456,3 +459,13 @@ func (s *SparkSchedulerExtender) deleteResourceReservation(namespace, name strin
 	background := metav1.DeletePropagationBackground
 	return s.resourceReservationClient.ResourceReservations(namespace).Delete(name, &metav1.DeleteOptions{PropagationPolicy: &background})
 }
+
+type set map[string]interface{}
```
onursatici: I think we use `map[string]bool` in other places, so we can use that for now, and no need to declare a new type as it is not re-used and is descriptive enough

internal/extender/resource.go
adambalogh: makes sense, I actually just extracted this into its own method, can you take a look again?

internal/extender/resource.go
```diff
@@ -198,22 +198,12 @@ func (s *SparkSchedulerExtender) fitEarlierDrivers(
...
 	availableNodes, err := s.nodeLister.List(labels.Set(driver.Spec.NodeSelector).AsSelector())
-	executorNodeNames := make([]string, 0, len(availableNodes))
-	for _, node := range availableNodes {
-		if !node.Spec.Unschedulable {
-			executorNodeNames = append(executorNodeNames, node.Name)
-		}
-	}
 	if err != nil {
-		return "", failureInternal, err
+		return "", failureFit, err
```
onursatici: should be `failureInternal`

adambalogh: ops, don't know why I changed that

PR #10, opened 2019-02-14
Cross AZ traffic metric

adambalogh:


internal/metrics/metrics.go
```diff
@@ -127,3 +130,58 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
 		schedulingRetryTime, sparkRoleTag, outcomeTag, s.instanceGroupTag, s.retryTag).Update(now.Sub(s.lastSeenTime).Nanoseconds())
 }
+
+func ReportCrossZoneMetric(ctx context.Context, driverNodeName string, executorNodeNames []string, nodes []*v1.Node) {
+	executorNodesSet := make(map[string]interface{})
+	for _, n := range executorNodeNames {
+		executorNodesSet[n] = nil
+	}
+
+	zonesCounter := make(map[string]int64)
```
onursatici: can use `int`

internal/metrics/metrics.go
```diff
@@ -127,3 +130,58 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+	for _, n := range nodes {
+		if _, ok := executorNodesSet[n.Name]; ok {
+			executorZone, err := getNodeZone(n)
+			if err != nil {
+				return
+			}
+			increment(zonesCounter, executorZone)
```
onursatici: you can use `zonesCounter[executorZone]++` and remove the increment method

internal/metrics/metrics.go
```diff
@@ -127,3 +130,58 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+}
+
+func increment(counter map[string]int64, zoneName string) {
+	if _, ok := counter[zoneName]; !ok {
+		counter[zoneName] = 0
+	}
+	counter[zoneName] += 1
+}
+
+func getNodeZone(node *v1.Node) (string, error) {
```
onursatici: because you don't handle the returned error, I think you can remove this function,
and above just do
```
driverZone, ok := node.Labels[nodeZoneLabel]
if !ok {
  // log warn here
}
```

internal/metrics/metrics.go
```diff
@@ -127,3 +130,58 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+	zone, ok := node.Labels[nodeZoneLabel]
+	if !ok {
+		return "", errors.New("zone label not found")
+	}
+	return zone, nil
+}
+
+func crossZoneTraffic(zonesCounter map[string]int64) int64 {
```
onursatici: can you add docs to this explaining the heuristics?

internal/metrics/metrics.go
adambalogh: ah nice, removed

internal/metrics/metrics.go
adambalogh: it's just the histogram expects int64, but I guess it's safe to convert int to int64 there

internal/metrics/metrics.go
adambalogh: yup, added a warn log

internal/metrics/metrics.go
adambalogh: added a comment

internal/metrics/metrics.go
```diff
@@ -127,3 +129,48 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
 	metrics.FromContext(ctx).Histogram(
 		schedulingRetryTime, sparkRoleTag, outcomeTag, s.instanceGroupTag, s.retryTag).Update(now.Sub(s.lastSeenTime).Nanoseconds())
 }
+
+// ReportCrossZoneMetric reports metric about cross AZ traffic between pods of a spark application
+func ReportCrossZoneMetric(ctx context.Context, driverNodeName string, executorNodeNames []string, nodes []*v1.Node) {
+	executorNodesSet := make(map[string]interface{})
```
onursatici: can we also use `map[string]bool` for consistency?

internal/metrics/metrics.go
```diff
@@ -127,3 +129,48 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+			zonesCounter[driverZone]++
+		}
+	}
+
+	metrics.FromContext(ctx).Histogram(crossAzTraffic).Update(int64(crossZoneTraffic(zonesCounter)))
+}
+
+// crossZoneTraffic calculates the total number of pairs of pods, where the 2 pods are in different zones.
+// A pair represents potential cross-zone traffic, which we want to avoid.
+func crossZoneTraffic(zonesCounter map[string]int) int {
```
onursatici: I wonder if we just do math here: It would be `C(total, 2) - C(num_az1, 2) - C(num_az2, 2)` for a two az cluster

onursatici: I don't think that would be  more efficient though, and given you explain what this does above, I think this is fine to keep

internal/metrics/metrics.go
```diff
@@ -127,3 +129,48 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+	var numPodsInDifferentZone int
```
onursatici: can also pass `len(executorNodes) + 1` to avoid the loop below

adambalogh: fixed

internal/metrics/metrics.go
```diff
@@ -46,6 +47,7 @@ const (
 	hostTagName           = "nodename"
 	lifecycleTagName      = "lifecycle"
 	sparkSchedulerName    = "spark-scheduler"
+	nodeZoneLabel         = "failure-domain.beta.kubernetes.io/zone"
```
ashrayjain: this should have come from a k8s constant ideally

internal/metrics/metrics.go
```diff
@@ -127,3 +129,44 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
 	metrics.FromContext(ctx).Histogram(
 		schedulingRetryTime, sparkRoleTag, outcomeTag, s.instanceGroupTag, s.retryTag).Update(now.Sub(s.lastSeenTime).Nanoseconds())
 }
+
+// ReportCrossZoneMetric reports metric about cross AZ traffic between pods of a spark application
+func ReportCrossZoneMetric(ctx context.Context, driverNodeName string, executorNodeNames []string, nodes []*v1.Node) {
+	executorNodesSet := make(map[string]bool)
```
ashrayjain: you should use a struct{} here instead of a bool

internal/metrics/metrics.go
```diff
@@ -127,3 +129,44 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+		executorNodesSet[n] = true
+	}
+
+	zonesCounter := make(map[string]int)
+	for _, n := range nodes {
+		if _, ok := executorNodesSet[n.Name]; ok {
+			executorZone, ok := n.Labels[nodeZoneLabel]
+			if !ok {
+				svc1log.FromContext(ctx).Warn("zone label not found for node", svc1log.SafeParam("nodeName", n.Name))
+				return
```
ashrayjain: why do we return here. does a single node not having the label mean we don't output this metric at all? that seems bad?

internal/metrics/metrics.go
adambalogh: we already have this convention of using this type for "sets" in scheduler

ashrayjain: ok, but just flagging that this isn't idomatic go :)

internal/metrics/metrics.go
onursatici: there are none

onursatici: The only one I saw was in kubelet code

ashrayjain: https://github.com/kubernetes/kubernetes/blob/a7966022d78bce63fa8212ea2519c99f49c2570e/staging/src/k8s.io/api/core/v1/well_known_labels.go#L21

ashrayjain: Ah, i see the issue. They only recently moved it to core/v1 https://github.com/kubernetes/kubernetes/pull/73616/files#diff-2ddc1f695cd4e7d5ef9a2d93148eeca7

:D

internal/metrics/metrics.go
onursatici: I think k8s codebase uses `bool`s for sets

ashrayjain: https://github.com/kubernetes/kubernetes/tree/v1.12.4/staging/src/k8s.io/apimachinery/pkg/util/sets

That uses struct{}. It's fine, just a matter a few bytes :D

onursatici: I don't know which one is the idiomatic way though. I see your example but there are far more bool sets in k8s/src then `struct{}`s
I agree `struct{}` has a better memory footprint, but if we care enough to change this, I think this should be a separate PR changing all set representations in the codebase. Happy to merge that though.

PR #13, opened 2019-02-20
Fix: release reservations for terminated pods

laflechejonathan:
When a spark application has executors that crash, those executors remain bound to their reservation. In the case where all executors are lost, this leads to a starved spark application which appears to be healthy, but is unable to request new executors.

This fix allows a new scheduling request to take a reservation if it is bound to a terminated pod.

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -420,14 +420,16 @@ func (s *SparkSchedulerExtender) findUnboundReservations(ctx context.Context, ex
 	if err != nil {
 		return nil, failureInternal, werror.Wrap(err, "failed to list pods")
 	}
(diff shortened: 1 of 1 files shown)


internal/extender/resource.go
```diff
@@ -439,6 +441,14 @@ func (s *SparkSchedulerExtender) findUnboundReservations(ctx context.Context, ex
 	return relocatableReservations, successRescheduled, nil
 }
 
+func isPodTerminated(pod *v1.Pod) bool {
+	allTerminated := false
```
onursatici: I think this should be initialised with `len(pod.Status.ContainerStatuses) > 0`. We can also early return if the container statuses is empty and set allTerminated to true and fold from there

PR #17, opened 2019-03-20
Debug log for large overhead containers

adambalogh:


internal/extender/overhead.go
```diff
@@ -146,10 +147,21 @@ func (o *OverheadComputer) compute(ctx context.Context) {
 	o.overheadLock.Unlock()
 }
 
-func podToResources(pod *v1.Pod) *resources.Resources {
+func podToResources(ctx context.Context, pod *v1.Pod) *resources.Resources {
+	oneCPU := resource.NewMilliQuantity(1000, resource.DecimalSI)
+	oneGiB := resource.NewQuantity(1*1024*1024*1024, resource.BinarySI)
```
onursatici: can these be constants?

internal/extender/overhead.go
```diff
@@ -146,10 +147,21 @@ func (o *OverheadComputer) compute(ctx context.Context) {
...
+
 	res := resources.Zero()
 	for _, c := range pod.Spec.Containers {
-		res.AddFromResourceList(c.Resources.Requests)
+		resourceRequests := c.Resources.Requests
+		if resourceRequests.Cpu().AsDec().Cmp(oneCPU.AsDec()) > 0 || resourceRequests.Memory().AsDec().Cmp(oneGiB.AsDec()) > 0 {
+			svc1log.FromContext(ctx).Debug("Container with no resource reservation has high resource requests",
+				svc1log.SafeParam("Pod", pod.Name),
```
onursatici: can this be `podName` to match other logs?

adambalogh: 👍 fixed

internal/extender/overhead.go
adambalogh: They can only be vars, not sure if that's better?

adambalogh: moved them outside

PR #28, opened 2019-04-18
Informer goroutine fatal logging

rkaram:
Wraps the scheduler informer and other background goroutines with the fatal logging handler to get more insights in case of panics

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -98,27 +98,28 @@ func NewExtender(
 	}
 }
 
-// Start is responsible for starting background goroutines for the SparkSchedulerExtender
+// Start is responsible for starting background tasks for the SparkSchedulerExtender
 func (s *SparkSchedulerExtender) Start(ctx context.Context) {
+	_ = wapp.RunWithFatalLogging(ctx, s.doStart)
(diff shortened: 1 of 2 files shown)


internal/extender/resource.go
```diff
@@ -97,25 +98,28 @@ func NewExtender(
...
 }
 
-// Start is responsible for starting background goroutines for the SparkSchedulerExtender
+// Start is responsible for starting background tasks for the SparkSchedulerExtender
 func (s *SparkSchedulerExtender) Start(ctx context.Context) {
+	_ = wapp.RunWithFatalLogging(ctx, s.doStart)
+}
+
+func (s *SparkSchedulerExtender) doStart(ctx context.Context) error {
 	if s.checkDemandCRDExists(ctx) {
```
onursatici: I think we want to keep the initial call to `s.checkDemandCRDExists(ctx)` sync, as we want to block checking if the crd exists to start scheduling

rkaram: Went back to something similar to what we had previously to explicitly show that we want this check to run once and then launch the rest of the logic in a goroutine.

PR #35, opened 2019-06-04
mark apps that exceed the cluster capacity with a custom pod condition

onursatici:


internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+// custom pod condition.
+type UnschedulablePodMarker struct {
+	nodeLister       corelisters.NodeLister
+	podLister        corelisters.PodLister
+	coreClient       corev1.CoreV1Interface
+	overheadComputer *OverheadComputer
+	binpacker        *Binpacker
+}
+
+// NewUnschedulablePodMarker creates a new UnschedulablePodMarker
```
laflechejonathan: nit - pointless comment

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+		podLister:        podLister,
+		coreClient:       coreClient,
+		overheadComputer: overheadComputer,
+		binpacker:        binpacker,
+	}
+}
+
+// Start starts periodic scanning for unschedulable pods
+func (u *UnschedulablePodMarker) Start(ctx context.Context) {
+	_ = wapp.RunWithFatalLogging(ctx, u.doStart)
```
laflechejonathan: any reason why you can't inline `doStart`?

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (u *UnschedulablePodMarker) doStart(ctx context.Context) error {
+	t := time.NewTicker(60 * time.Second)
```
laflechejonathan: pull the interval into a named constant for readability. Also, what's your reasoning for choosing 1 minute as the poll interval?

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (u *UnschedulablePodMarker) mark(ctx context.Context) {
+	pods, err := u.podLister.List(labels.Everything())
+	if err != nil {
+		svc1log.FromContext(ctx).Error("failed to list pods", svc1log.Stacktrace(err))
+		return
+	}
+	now := time.Now()
+	for _, pod := range pods {
+		if pod.Spec.SchedulerName == SparkSchedulerName && // TODO: constant
```
laflechejonathan: not clear what this TODO means?

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+			len(pod.Spec.NodeName) == 0 &&
+			pod.DeletionTimestamp == nil &&
+			pod.Labels[SparkRoleLabel] == Driver &&
+			pod.CreationTimestamp.Time.Add(10*time.Minute).Before(now) {
```
laflechejonathan: move 10 min. to constant alongside the poll interval, since changing one likely requires thinking about the other.

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+					svc1log.SafeParam("podName", pod.Name),
+					svc1log.SafeParam("podNamespace", pod.Namespace),
+					svc1log.Stacktrace(err))
+			}
+		}
+
+	}
+}
+
+func (u *UnschedulablePodMarker) markPod(ctx context.Context, driver *v1.Pod) error {
```
laflechejonathan: I would factor this into two functions for readability:
```
exceedsCapacity, err = doesApplicationExceedClusterCapacity(driverPod);
markClusterCapacityPodCondition(pod, exceedsCapacity)
```

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+	nodes, err := u.nodeLister.List(labels.Set(driver.Spec.NodeSelector).AsSelector())
+	if err != nil {
+		return err
+	}
+	nodeNames := make([]string, 0, len(nodes))
```
laflechejonathan: nit - I'm guessing all this boilerplate for getting a list of node names is copied from elsewhere, can we move to a utility function?

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,135 @@
...
+	applicationResources, err := sparkResources(ctx, driver)
+	if err != nil {
+		return err
+	}
+	_, _, hasCapacity := u.binpacker.BinpackFunc(
+		ctx,
+		applicationResources.driverResources,
+		applicationResources.executorResources,
+		applicationResources.executorCount,
+		nodeNames, nodeNames, availableResources)
```
laflechejonathan: nit - is this a gofmt thing? it seems weird that the last 3 params are on the same line

internal/extender/unschedulablepods.go
laflechejonathan: I suppose there should be a function `resources.availableResources()` that hides the node listing + overhead computation

internal/extender/unschedulablepods.go
laflechejonathan: small thing, but across your comments you do something like:

`{functionName} does X`

I'd shorten this to:
`does X`

to avoid things like:
`Start starts...`

laflechejonathan: @ashrayjain tells me this is a linting thing, so be it.

internal/extender/unschedulablepods.go
```diff
@@ -0,0 +1,155 @@
...
+		if pod.Spec.SchedulerName == SparkSchedulerName &&
+			len(pod.Spec.NodeName) == 0 &&
+			pod.DeletionTimestamp == nil &&
+			pod.Labels[SparkRoleLabel] == Driver &&
+			pod.CreationTimestamp.Time.Add(unschedulableInClusterThreshold).Before(now) {
+
+			exceedsCapacity, err := u.doesPodExceedClusterCapacity(ctx, pod)
+			if err != nil {
+				svc1log.FromContext(ctx).Error("failed to check if pod was unschedulable",
+					svc1log.SafeParam("podName", pod.Name),
```
onursatici: you can capture `podName` and `podNamespace` using `ctx = svc1log.WithLoggerParams(ctx, params...)`

internal/extender/unschedulablepods.go
onursatici: they do slightly different things, we don't add resource reservation usage here

PR #41, opened 2019-07-18
cache & async clients

onursatici:


internal/cache/async.go
```diff
@@ -0,0 +1,155 @@
...
+
+type object interface {
+	metav1.Object
+	runtime.Object
+}
+
+type asyncClient struct {
+	client             rest.Interface
+	resourceName       string
+	emptyObjectCreator func() object
```
laflechejonathan: seems weird

internal/cache/async.go
```diff
@@ -0,0 +1,155 @@
...
+	queue              store.ShardedUniqueQueue
+	objectStore        store.ObjectStore
+}
+
+func (as *asyncClient) Run(ctx context.Context) {
+	for _, q := range as.queue.GetConsumers() {
```
laflechejonathan: nit - would prefer you name `this` from `as` to `ac`, only because I read it like "as queue" "as runWorker" which I find confusing

internal/cache/async.go
```diff
@@ -0,0 +1,155 @@
...
+func isRetryableError(ctx context.Context, err error) bool {
+	isRetryable := errors.IsServerTimeout(err) || errors.IsServiceUnavailable(err) ||
+		errors.IsTooManyRequests(err) || errors.IsTimeout(err)
+	if isRetryable {
+		svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
+	}
+	return isRetryable
+}
+
+func logAndIgnore(ctx context.Context, err error) {
```
laflechejonathan: rename to `logNonRetryableError`

internal/cache/async.go
```diff
@@ -0,0 +1,155 @@
...
+	default:
+		logAndIgnore(ctx, err)
+	}
+}
+
+func isRetryableError(ctx context.Context, err error) bool {
+	isRetryable := errors.IsServerTimeout(err) || errors.IsServiceUnavailable(err) ||
+		errors.IsTooManyRequests(err) || errors.IsTimeout(err)
+	if isRetryable {
+		svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
```
laflechejonathan: nit - I'm not a fan of logging `will do X` in a function that just returns a boolean, who knows if the caller will do the right thing.

Instead, maybe factor out `as.queue.AddIfAbsent(...)` as a function called `addTaskToRetry(...)` and move that log statement there?

internal/cache/async.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		as.objectStore.OverrideResourceVersionIfNewer(ctx, obj)
+	}
+	switch {
+	case err == nil:
+		as.objectStore.OverrideResourceVersionIfNewer(ctx, obj)
+	case isRetryableError(ctx, err):
+		as.queue.AddIfAbsent(store.UpdateRequest(obj))
+	case errors.IsConflict(err):
+		svc1log.FromContext(ctx).Warn("got conflict, will retry", svc1log.Stacktrace(err))
```
laflechejonathan: How convinced are you that this is the right behaviour?

internal/cache/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	)
+	return c
+}
+
+func (c *cache) Create(obj metav1.Object) bool {
+	created := c.store.PutIfAbsent(obj)
+	if !created {
+		return false
+	}
+	c.queue.AddIfAbsent(store.CreateRequest(obj))
```
laflechejonathan: do you care whether `AddIfAbsent` actually adds to the queue?

internal/cache/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	return true
+}
+
+func (c *cache) Get(namespace, name string) (metav1.Object, bool) {
+	return c.store.Get(store.Key{Namespace: namespace, Name: name})
+}
+
+func (c *cache) Update(obj metav1.Object) bool {
```
laflechejonathan: question - in here and in Create, would it be more idiomatic to return an error, which is nil in the happy case?

internal/cache/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	typedObject, ok := obj.(metav1.Object)
+	if !ok {
+		svc1log.FromContext(ctx).Warn("failed to parse object")
+		return
+	}
+	c.store.OverrideResourceVersionIfNewer(ctx, typedObject)
+}
+
+func (c *cache) onObjUpdate(oldObj interface{}, newObj interface{}) {
+	ctx := context.Background()
```
laflechejonathan: this is entirely the same as onObjAdd - can we make one delegate to the other (probably add delegates to update)

internal/cache/demands.go
```diff
@@ -0,0 +1,83 @@
...
+) (*DemandCache, error) {
+	ds, err := demandInformer.Lister().List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	objectStore := store.NewStore()
+	for _, d := range ds {
+		objectStore.Put(d)
+	}
+	queue := store.NewShardedUniqueQueue(5)
```
laflechejonathan: move magic number to constant, maybe with a comment that explains how that number should be chosen?

internal/cache/store/queue.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"hash/fnv"
+	"sync"
+)
+
+// ShardedUniqueQueue is a queue of write requests
+// for objects. It compacts consecutive create and update
+// requests, provides a slice of channels for consumers.
+// Requests consumed from these channels are mutually
```
laflechejonathan: What does this mean to you:
"Requests consumed from these channels are mutually exclusive"

I'm not sure I understand.

internal/cache/store/queue.go
```diff
@@ -0,0 +1,95 @@
...
+type ShardedUniqueQueue interface {
+	AddIfAbsent(Request)
+	GetConsumers() []<-chan func() Request
+}
+
+// NewShardedUniqueQueue creates a sharded queue of write requests
+func NewShardedUniqueQueue(buckets int) ShardedUniqueQueue {
+	queues := make([]chan func() Request, 0, buckets)
+	for i := 0; i < buckets; i++ {
+		queues = append(queues, make(chan func() Request, 100))
```
laflechejonathan: What is 100, can we extract that magic number somewhere?

internal/cache/store/queue.go
```diff
@@ -0,0 +1,95 @@
...
+	queues   []chan func() Request
+	inflight map[Key]bool
+	lock     sync.Mutex
+}
+
+// AddIfAbsent puts a request to the queue if it is absent,
+// or if it is a delete request. Deletes are not compacted
+// to prior update or create requests, so all input objects
+// can be created
+func (q *shardedUniqueQueue) AddIfAbsent(r Request) {
```
laflechejonathan: this code is interesting and weird! I have many questions, let's discuss.

internal/cache/store/request.go
```diff
@@ -0,0 +1,71 @@
...
+// KeyOf creates a key from an object
+func KeyOf(obj metav1.Object) Key {
+	return Key{obj.GetNamespace(), obj.GetName()}
+}
+
+// RequestType denotes the type of write request
+type RequestType int
+
+const (
+	// CreateRequestType denotes create requests
```
laflechejonathan: I think these comments are unnecessary

internal/cache/store/request.go
```diff
@@ -0,0 +1,71 @@
...
+	DeleteRequestType RequestType = 2
+)
+
+// Request is a write request for an object
+type Request struct {
+	Key  Key
+	Type RequestType
+}
+
+// CreateRequest creates a create request for an object
```
laflechejonathan: same for these

internal/cache/store/store.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewStore creates an empty store
+func NewStore() ObjectStore {
+	return &objectStore{
+		store: make(map[Key]metav1.Object),
+	}
+}
+
+func (s *objectStore) Put(obj metav1.Object) {
+	s.lock.Lock()
```
laflechejonathan: I'm curious whether the mutex approach is idiomatic for go?

I imagine you can also do this with channels and a single goroutine that maintains the map, would that be more clear?

internal/cache/store/store.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (s *objectStore) Get(key Key) (metav1.Object, bool) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	obj, exists := s.store[key]
+	return obj, exists
+}
+
+func (s *objectStore) Delete(key Key) {
+	_, ok := s.Get(key)
```
laflechejonathan: Is the get meant as an optimisation or a correctness thing? If the latter, seems there's a critical section where you can miss a racing delete?

internal/cache/async.go
onursatici: There is no interface both metav1.Object and runtime.Object satisfy unfortunately

internal/cache/async.go
onursatici: Actually I am going to get the latest and update again here to force the update go through, I think retrying indefinitely is error prone

internal/cache/cache.go
onursatici: I think it has to, it blocks until the request is sent to the channel

internal/cache/cache.go
onursatici: Yea that makes sense for these, it is not apparent what does a return value of false mean from an Update call

internal/cache/store/queue.go
onursatici: It means no two requests for the same object would end up in different consumers, will reword

internal/cache/store/queue.go
onursatici: We can, it is the number of in flight write requests we can tolerate until we start blocking for additional write requests

internal/cache/store/queue.go
onursatici: The idea is to add a write request only if there are no other write requests in the queue. I have gone with excluding delete requests from this, because I want each object to appear in etcd at one point. Example would be a create call and a delete call while the create call was in flight, if you compact the two you can get away with doing nothing, but etcd will be unaware of that object has existed in our cache at one point. So mainly I want this for increased visibility

internal/cache/store/store.go
onursatici: The former, I don’t wanna hold the write lock for delete calls on non existent objects

internal/cache/store/store.go
onursatici: If you go with that way you will need a goroutine for handling map updates, I think in this case where you mainly want a concurrent map you shouldn’t overuse channels

internal/cache/store/request.go
onursatici: golint complains afaik, I can double check

internal/cache/async.go
onursatici: now using an interface to abstract types out from the clients

internal/cache/async.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	default:
+		logNonRetryableError(ctx, err)
+	}
+}
+
+func (ac *asyncClient) doDelete(ctx context.Context, key store.Key) {
+	err := ac.client.Delete(key.Namespace, key.Name)
+	if err != nil {
+		ac.objectStore.Delete(key)
```
rkaram: This might not be needed

internal/cache/async.go
```diff
@@ -0,0 +1,147 @@
...
+		svc1log.FromContext(ctx).Info("Ignoring request for deleted object")
+		return
+	}
+	result, err := ac.client.Create(obj)
+	switch {
+	case err == nil:
+		ac.objectStore.OverrideResourceVersionIfNewer(ctx, result)
+	case isRetryableError(err):
+		svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
+		ac.queue.AddIfAbsent(store.CreateRequest(obj))
```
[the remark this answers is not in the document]
onursatici: there is the qps and burst config on the client, so you will get throttling there

internal/cache/async.go
```diff
@@ -0,0 +1,147 @@
...
+	err := ac.client.Delete(key.Namespace, key.Name)
+	if err != nil {
+		ac.objectStore.Delete(key)
+	}
+	switch {
+	case err == nil:
+		ac.objectStore.Delete(key)
+	case isRetryableError(err):
+		svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
+		ac.queue.AddIfAbsent(store.Request{Key: key, Type: store.DeleteRequestType})
```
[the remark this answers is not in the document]
onursatici: Yea but it is internal and is not a pointer so I didn't want to create an extra copy just for a slight readability improvement, I think construction tells what it is doing

internal/cache/async.go
onursatici: good catch, deleted

onursatici:
@rkaram yea will add metrics, but was thinking about adding them in a follow up pr, so my ordering would be:
- this
- leader failover
- metrics

PR #43, opened 2019-07-31
cache reconciliation on leader failover

onursatici:


internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+		return nil, nil, err
+	}
+
+	nodes, nodesFound := r.orderedNodes[instanceGroup]
+	availableResources, resourcesFound := r.availableResources[instanceGroup]
+	if !nodesFound || !resourcesFound {
+		return nil, nil, werror.Error("instance group not found", werror.SafeParam("instanceGroup", instanceGroup))
+	}
+
+	executorCountToReserve := applicationResources.executorCount - len(executors)
```
rkaram: Don't you want to create reservations for stale executors instead?

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+		// the driver is stale, a new resource reservation object needs to be created
+		instanceGroup := sp.driver.Spec.NodeSelector[instanceGroupNodeSelector]
+		newRR, reservedResources, err := r.constructResourceReservation(ctx, sp.driver, sp.executors, instanceGroup)
+		if err != nil {
+			svc1log.FromContext(ctx).Error("failed to construct resource reservation", svc1log.Stacktrace(err))
+			return
+		}
+		err = r.resourceReservations.Create(newRR)
+		if err != nil {
+			logRR(ctx, "resource reservation already exists, force updating", sp.driver.Namespace, sp.driver.Labels[SparkAppIDLabel])
```
rkaram: Just curious in which scenario you're expecting this to happen...

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func (r *reconciliator) deleteDemandIfExists(namespace, name string) {
+	_, ok := r.demands.Get(namespace, name)
+	if ok {
+		r.demands.Delete(namespace, name)
+	}
+}
+
+func sparkPodsByAppID(
```
rkaram: nit, but I think this should either have a name or documentation that clarifies it drops already reserved pods.

internal/logging.go
```diff
@@ -35,6 +36,14 @@ func DemandSafeParams(demandName string, demandNamespace string) map[string]inte
 	}
 }
 
+// ResourceReservationSafeParamsFromObj gets the safe params for a resource reservation object
+func ResourceReservationSafeParamsFromObj(rr *v1beta1.ResourceReservation) map[string]interface{} {
```
rkaram: This does not seem to be used

internal/cache/safedemands.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (sdc *SafeDemandCache) IsLoaded() bool {
+	return sdc.demandCRDInitialized.Load()
+}
+
+// Create enqueues a creation request and puts the object into the store
+func (sdc *SafeDemandCache) Create(rr *demandapi.Demand) error {
+	if !sdc.demandCRDInitialized.Load() {
+		return nil
```
laflechejonathan: should return an error

internal/extender/demand.go
```diff
@@ -60,7 +56,7 @@ func (s *SparkSchedulerExtender) updatePodStatus(ctx context.Context, pod *v1.Po
 }
 
 func (s *SparkSchedulerExtender) createDemandForExecutor(ctx context.Context, executorPod *v1.Pod, executorResources *resources.Resources) {
-	if !s.demandCRDInitialized.Load() {
+	if !s.demands.IsLoaded() {
```
laflechejonathan: s/loaded/exists/ or registered

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	corelisters "k8s.io/client-go/listers/core/v1"
+)
+
+// SyncResourceReservationsAndDemands gets all resource reservations and pods,
+// finds all pods that do not have a claimed spot in their resource reservation object,
+// groups them by spark application id, and patched the resource reservation objects so
```
laflechejonathan: *patches

internal/extender/failover.go
laflechejonathan: something like `getUnreservedSparkPodsBySparkID`

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+	// if the driver is nil it already has an associated reservation, get the resource
+	// reservation object and update it so it has reservations for each stale executor
+	if sp.driver == nil && len(sp.executors) > 0 {
+		exec := sp.executors[0]
+		rr, ok := r.resourceReservations.Get(exec.Namespace, exec.Labels[SparkAppIDLabel])
+		if !ok {
+			logRR(ctx, "resource reservation deleted, ignoring", exec.Namespace, exec.Labels[SparkAppIDLabel])
+			return
+		}
+		err := r.patchResourceReservation(sp.executors, rr.DeepCopy())
```
laflechejonathan: not clear to me why you need a Deep Copy

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+	availableResources   map[string]resources.NodeGroupResources
+	orderedNodes         map[string][]*v1.Node
+}
+
+func (r *reconciliator) syncResourceReservation(ctx context.Context, sp *sparkPods) {
+	// if the driver is nil it already has an associated reservation, get the resource
+	// reservation object and update it so it has reservations for each stale executor
+	if sp.driver == nil && len(sp.executors) > 0 {
+		exec := sp.executors[0]
+		rr, ok := r.resourceReservations.Get(exec.Namespace, exec.Labels[SparkAppIDLabel])
```
laflechejonathan: I think your code will be cleaner if `sparkPods` had a SparkAppId field.

internal/extender/failover.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	// recompute overhead to account for newly created resource reservations
+	overheadComputer.compute(ctx)
+	return nil
+}
+
+// sparkPods is a collection of stale state, it is comprised of
+// pods from a spark application that do not have a claimed resource reservation,
+// and the last known state of the resource reservation object
+type sparkPods struct {
```
laflechejonathan: rename to something like `inconsistentSparkApplication`

laflechejonathan: or actually, do this:
```golang
type sparkApplication struct {
    sparkAppID string
    inconsistentDriver *v1.Pod
    inconsistentExecutors []*v1.Pod
}```

File: internal/extender/failover.go [hunk 11]
```diff
@@ -0,0 +1,296 @@
...
+				logRR(ctx, "resource reservation deleted, ignoring", sp.driver.Namespace, sp.driver.Labels[SparkAppIDLabel])
+				return
+			}
+		}
+		r.availableResources[instanceGroup].Sub(reservedResources)
+	}
+}
+
+func (r *reconciliator) syncDemand(ctx context.Context, sp *sparkPods) {
+	// TODO check if demand crd exists
```
laflechejonathan: rm

internal/extender/failover.go
onursatici: Not expecting this to happen, but want to guard against two extenders starting up. What I will do is to make sure we run only on the leader on startup

internal/extender/failover.go
onursatici: I think the wording is off here, these executors do not exist yet so we need to reserve spaces for them, as opposed to `executors` here that exist but do not have the corresponding resource reservation object

internal/extender/failover.go
onursatici: you can't change objects you got from the informer / cache

PR #46, opened 2019-08-02
cache metrics

onursatici:


internal/cache/safedemands.go
```diff
@@ -150,3 +150,14 @@ func (sdc *SafeDemandCache) Get(namespace, name string) (*demandapi.Demand, bool
 	}
 	return sdc.DemandCache.Get(namespace, name)
 }
+
+// CachedElements returns the number of elements in the cache
+func (sdc *SafeDemandCache) CachedElements() int {
```
rkaram: Name should indicate this is a count

internal/metrics/cache.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+			rrsCached := c.resourceReservations.List()
+			metrics.FromContext(ctx).Gauge(cachedObjectCount, rrTag, listerTag).Update(int64(len(rrs)))
+			metrics.FromContext(ctx).Gauge(cachedObjectCount, rrTag, cacheTag).Update(int64(len(rrsCached)))
+			for idx, queueLength := range c.resourceReservations.InflightQueueLengths() {
+				metrics.FromContext(ctx).Gauge(inflightRequestCount, rrTag, QueueIndexTag(ctx, idx)).Update(int64(queueLength))
+			}
+
+			if !c.demands.CRDExists() {
+				continue
```
laflechejonathan: nit - seems like you should invert that condition and move the next bunch of statements into the condition, maybe worth factoring out as:

```
emitResourceReservationMetrics()
if c.demands.CRDExists() {
    emitDemandMetrics()
}
```

PR #47, opened 2019-08-02
reconcile on requests

onursatici:


internal/extender/resource.go
```diff
@@ -35,14 +36,15 @@ import (
...
-	successAlreadyBound  = "success-already-bound"
+	failureUnbound         = "failure-unbound"
+	failureInternal        = "failure-internal"
+	failureFit             = "failure-fit"
+	failureEarlierDriver   = "failure-earlier-driver"
+	failureNonSparkPod     = "failure-non-spark-pod"
+	success                = "success"
+	successRescheduled     = "success-rescheduled"
+	successAlreadyBound    = "success-already-bound"
+	leaderElectionInterval = 15 * time.Second
```
laflechejonathan: add comment explaining how this time period was chosen

internal/extender/resource.go
```diff
@@ -120,6 +131,16 @@ func failWithMessage(ctx context.Context, args schedulerapi.ExtenderArgs, messag
 	return &schedulerapi.ExtenderFilterResult{FailedNodes: failedNodes}
 }
 
+func (s *SparkSchedulerExtender) reconcileIfNeeded(ctx context.Context) error {
+	now := time.Now()
+	var err error
+	if s.lastRequest.Add(leaderElectionInterval).After(now) {
+		err = s.syncResourceReservationsAndDemands(ctx)
+	}
+	s.lastRequest = now
```
laflechejonathan: I think you only want to set this if err != nil

onursatici: done

PR #50, opened 2019-08-05
factor test utils out

laflechejonathan:


internal/extender/extendertest/extender_test_utils.go
```diff
@@ -0,0 +1,254 @@
...
+		fakeAPIExtensionsClient,
+		isFIFO,
+		binpacker,
+		overheadComputer,
+	)
+
+	return &Harness{
+		Extender:  extender,
+		PodStore:  podInformer.Informer().GetStore(),
+		NodeStore: nodeInformer.Informer().GetStore(),
```
onursatici: can you also add resource reservation store here?

PR #55, opened 2019-08-07
log on deletions, namespace not found

onursatici:


internal/cache/cache.go
```diff
@@ -112,3 +114,11 @@ func (c *cache) tryOverrideResourceVersion(ctx context.Context, obj interface{})
 	}
 	c.store.OverrideResourceVersionIfNewer(ctx, typedObject)
 }
+
+// ObjectSafeParams returns safe logging params for a name and a namespace
+func ObjectSafeParams(name, namespace string) svc1log.Param {
```
rkaram: Why not take the object and select needed params in here?

rkaram: Why not take the object and select needed params in here?

onursatici: I think one of the callers of this only has a `Key` object, instead of the full `metav1.Object`

PR #62, opened 2019-09-11
Fix: Add test for unschedulable pod marker

laflechejonathan:


internal/extender/unschedulablepods.go
```diff
@@ -122,6 +123,11 @@ func (u *UnschedulablePodMarker) doesPodExceedClusterCapacity(ctx context.Contex
 	for _, node := range nodes {
 		nodeNames = append(nodeNames, node.Name)
 	}
+	if len(nodeNames) == 0 {
+		svc1log.FromContext(ctx).Warn("could not find any nodes matching pod selectors, possible misconfiguration.",
```
rkaram: small nit - it could be a case of a pending scale up as well for example, so maybe we should mention this as well since it is more likely than a misconfiguration...

laflechejonathan: Fair enough, might make it info as well

PR #65, opened 2019-09-18
[Dynamic Allocation] Adds metrics for SoftReservationStore

rkaram:
Adds metric reporting on reservation and total executor counts for the SoftReservationStore


rkaram:
@onursatici I definitely think it would be useful to have the metrics tagged by instance group; however, we don't currently store any instance group or pod information where we can extract this from in the `SoftReservationStore`. 
We can add the instance group of the driver when creating the soft reservation, however the executors could technically have different instance group labels, and we would need to store that as well instead of the `Reservation` struct only.
The alternative is to get the information by cross referencing the executor names with a pod lister, but then some information could be missing for some executors, and I would prefer not going with this approach.

It shouldn't be overly complicated, but would have to add some additional structs, and I want to see if you think it's worth it at this moment, or we should leave it to a later time?

onursatici:
Yes I think it makes sense to not do that on this PR. One more thing to possibly add here is, in `emitSoftReservationMetrics`, should we also get the actual extra executor count from the pod lister and reservations lister, so it would be trivial to alert if they deviate

PR #66, opened 2019-09-20
Fix: Make instance group label configurable

laflechejonathan:


internal/extender/demand.go
```diff
@@ -135,8 +133,7 @@ func newDemand(pod *v1.Pod, units []demandapi.DemandUnit) (*demandapi.Demand, er
 			Name:      demandName,
 			Namespace: pod.Namespace,
 			Labels: map[string]string{
-				SparkAppIDLabel:    appID,
-				instanceGroupLabel: instanceGroup,
+				SparkAppIDLabel: appID,
```
laflechejonathan: @ashrayjain I am assuming that this instanceGroup label that was set by scheduler on the demand object was not actually needed, and instead the instanceGroup on the DemandSpec below is what matters - does that match your understanding?

ashrayjain: yes, that is correct. we use the value from the spec and don't expect the label to be present.

config/config.go
```diff
@@ -20,13 +20,14 @@ import (
...
-	Burst          int     `yaml:"burst,omitempty"`
-	BinpackAlgo    string  `yaml:"binpack,omitempty"`
+	config.Install     `yaml:",inline"`
+	config.Runtime     `yaml:",inline"`
+	Kubeconfig         string  `yaml:"kube-config,omitempty"`
+	FIFO               bool    `yaml:"fifo,omitempty"`
+	QPS                float32 `yaml:"qps,omitempty"`
+	Burst              int     `yaml:"burst,omitempty"`
+	BinpackAlgo        string  `yaml:"binpack,omitempty"`
+	InstanceGroupLabel string  `yaml:"InstanceGroupLabel,omitempty"`
```
rkaram: Two things here:
1. to avoid a breaking change, we should set this to a default `resource_channel` value
2. (nit) we should have this key in the same case as the rest of the config rather than Pascal Case, for consistency

laflechejonathan:
Thanks for the review @rkaram, all good points. For the change to `nodeSelector`, I think you're correct that the changes shouldn't be bundled, but for expediency I'd like to leave it in.

I'm confident that we don't support FIFO use cases with drivers that have heterogenous node selectors, speaking to @onursatici this was just an accident of implementation.

Alternatively, I can pull this change into its own PR which we can merge first? let me know if you'd like me to do that.

internal/extender/sparkpods.go
```diff
@@ -84,7 +86,7 @@ func filterToEarliestAndSort(driver *v1.Pod, allDrivers []*v1.Pod) []*v1.Pod {
 		// add only unscheduled drivers with the same instance group and targeted to the same scheduler
 		if len(p.Spec.NodeName) == 0 &&
 			p.Spec.SchedulerName == driver.Spec.SchedulerName &&
-			p.Spec.NodeSelector[instanceGroupNodeSelector] == driver.Spec.NodeSelector[instanceGroupNodeSelector] &&
+			reflect.DeepEqual(p.Spec.NodeSelector, driver.Spec.NodeSelector) &&
```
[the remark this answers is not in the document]
onursatici: Yea thinking about this a bit more, I think we should keep the instanceGroupLabel as the construct for queues in FIFO. 
Thinking about possible future work where we might impose additional constraints to some of the applications, that should not change the queue the pod would be in regarding FIFO. If we consider the value of only a single label in FIFO queues, then we guarantee a mutually exclusive set of nodes per queue.

rkaram:
Merged 🎉
Note: we probably want to also remove `InstanceGroupLabel` from `k8s-spark-scheduler-lib` as it doesn't seem to be used https://github.com/palantir/k8s-spark-scheduler-lib/blob/master/pkg/apis/sparkscheduler/v1beta1/types.go#L24

PR #67, opened 2019-09-23
[Fix] Check if resource reservation already exists before fitting driver

rkaram:
Adds a check for existing resource reservation for a driver before we attempt to schedule it in case we get a double submission from `kube-scheduler`.
We currently try to binpack a second time and only if the driver fits do we return the previous resource reservation's node, but we create a duplicate demand if the driver does not fit. This PR makes the behavior consistent between the two cases, and reduces the chances of double creating demands.

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -197,12 +197,12 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
 		driverReservedNode := rr.Spec.Reservations["driver"].Node
 		for _, node := range nodeNames {
 			if driverReservedNode == node {
-				svc1log.FromContext(ctx).Info("Received request to schedule driver which already has a reservation. Returning previously reserved node...",
+				svc1log.FromContext(ctx).Info("Received request to schedule driver which already has a reservation. Returning previously reserved node.",
 					svc1log.SafeParam("driverReservedNode", driverReservedNode))
 				return driverReservedNode, success, nil
 			}
 		}
-		svc1log.FromContext(ctx).Warn("Received request to schedule driver which already has a reservation, but previously reserved node is not in list of nodes. Returning previously reserved node anyway...",
+		svc1log.FromContext(ctx).Warn("Received request to schedule driver which already has a reservation, but previously reserved node is not in list of nodes. Returning previously reserved node anyway.",
 			svc1log.SafeParam("driverReservedNode", driverReservedNode),
 			svc1log.SafeParam("nodeNames", nodeNames))
 		return driverReservedNode, success, nil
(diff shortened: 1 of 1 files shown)


internal/extender/resource.go
```diff
@@ -193,6 +193,21 @@ func (s *SparkSchedulerExtender) fitEarlierDrivers(
 }
 
 func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v1.Pod, nodeNames []string) (string, string, error) {
+	if rr, ok := s.resourceReservations.Get(driver.Namespace, driver.Labels[SparkAppIDLabel]); ok {
+		driverReservedNode := rr.Spec.Reservations["driver"].Node
+		for _, node := range nodeNames {
+			if driverReservedNode == node {
+				svc1log.FromContext(ctx).Info("Received request to schedule driver which already has a reservation. Returning previously reserved node...",
+					svc1log.SafeParam("namespace", driver.Namespace),
```
onursatici: you can remove namespace, as it is already in params of this ctx

internal/extender/resource.go
```diff
@@ -193,6 +193,21 @@ func (s *SparkSchedulerExtender) fitEarlierDrivers(
...
+					svc1log.SafeParam("driverReservedNode", driverReservedNode))
+				return driverReservedNode, success, nil
+			}
+		}
+		svc1log.FromContext(ctx).Warn("Received request to schedule driver which already has a reservation, but previously reserved node is not in list of nodes",
```
onursatici: I think we should log Warn, but still return the driverReservedNode here to preserve the previous behaviour

rkaram: Addressed for consistency with previous behavior, but what is `kube-scheduler`'s behavior in this case?

internal/extender/resource.go
```diff
@@ -193,6 +193,20 @@ func (s *SparkSchedulerExtender) fitEarlierDrivers(
...
+	if rr, ok := s.resourceReservations.Get(driver.Namespace, driver.Labels[SparkAppIDLabel]); ok {
+		driverReservedNode := rr.Spec.Reservations["driver"].Node
+		for _, node := range nodeNames {
+			if driverReservedNode == node {
+				svc1log.FromContext(ctx).Info("Received request to schedule driver which already has a reservation. Returning previously reserved node...",
+					svc1log.SafeParam("driverReservedNode", driverReservedNode))
+				return driverReservedNode, success, nil
+			}
+		}
+		svc1log.FromContext(ctx).Warn("Received request to schedule driver which already has a reservation, but previously reserved node is not in list of nodes. Returning previously reserved node anyway...",
```
laflechejonathan: nit - remove "..."

internal/extender/resource.go
onursatici: I think currently it doesn't check filter result, so extenders can return nodes that were not part of the input list.
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/core/generic_scheduler.go#L530-L547

PR #68, opened 2019-09-25
[Fix] Call informer factory Start() only after creating the needed informers

rkaram:
The `sharedInformerFactory.Start()` method loops over its internal `informers` map when called and calls `Run()` on each informer that hasn't been started yet. An informer is effectively created and added to the factory's `informers` map when its `Informer()` or `Lister()` methods are invoked and not on initialization.

This PR makes sure we create the needed informers before starting the factories so we avoid entering into race conditions when `Start()` is called in goroutines, or failures when called synchronously but in the reverse order.


onursatici:
thanks @rkaram nice find!

PR #74, opened 2019-10-15
Fixes bug in failover where we were reading the outdated resource reservation object

rkaram:
Fixes a bug in failover where we were still reading the outdated resource reservation object when checking which executors need to have soft reservations.

### DIFF
--- a/internal/extender/failover.go
+++ b/internal/extender/failover.go
@@ -107,8 +107,13 @@ func (r *reconciler) syncResourceReservations(ctx context.Context, sp *sparkPods
 			return
 		}
 
-		podsWithRR := make(map[string]bool, len(rr.Status.Pods))
-		for _, podName := range rr.Status.Pods {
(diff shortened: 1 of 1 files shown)


internal/extender/failover.go
```diff
@@ -107,8 +107,13 @@ func (r *reconciler) syncResourceReservations(ctx context.Context, sp *sparkPods
 			return
 		}
 
-		podsWithRR := make(map[string]bool, len(rr.Status.Pods))
-		for _, podName := range rr.Status.Pods {
+		newRR, ok := r.resourceReservations.Get(exec.Namespace, sp.appID)
```
onursatici: I think you can return the patched resource reservation from the call to `patchResourceReservation` above, that way you don't have to look up from the cache here

rkaram: 👍 Done

PR #75, opened 2019-10-17
Add event logs on scheduling and demand interactions

ashrayjain:


internal/extender/extendertest/extender_test_utils.go
```diff
@@ -102,9 +103,6 @@ func NewTestExtender(objects ...runtime.Object) (*Harness, error) {
 		fakeAPIExtensionsClient,
 		fakeSchedulerClient.ScalerV1alpha1(),
 	)
-	if err != nil {
```
ashrayjain: this was dead code.

rkaram:
I personally think it would be cleaner if these were methods in the event package rather than fully inline. Something like `events.emitDemandCreated(*demandapi.Demand)`.
But overall good idea, we should also add event logs for some other important events as well

rkaram:
lgtm 👍

PR #76, opened 2019-10-19
Prioritize hosts by current usage rather than just lifetime

j-baker:
We kill our hosts regularly, and so we would like to schedule drivers so
that they heuristically get the longest remaining lifetime. For this
reason, we currently sort our hosts purely in terms of their remaining
host lifetime.

In practice, this is demonstrably wasteful. Consider the case where we
have hosts A and B, where A has a single pod running on it and we've
just booted up host B and so it has no pods running on it. An arriving
job will be scheduled onto host B, leading to poor usage. We would
prefer to schedule the arriving job onto A as well, and be able to kill
B.

This PR takes current usage into account. First, it attempts to fill up
old nodes, regardless of their usage. For young nodes (the ones where
suitability for scheduling drivers is immaterial), we instead schedule
first based on memory (prefer the hosts with the least remaining memory)
and then CPU (prefer the hosts with the least remaining CPU).

The goal is that instead of having an even spread of partially used
hosts, we'd be able to partition into totally used hosts and mostly
unused hosts.

### DIFF
--- a/internal/extender/hostordering_test.go
+++ b/internal/extender/hostordering_test.go
@@ -1,0 +1,60 @@
+package extender
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	"testing"
+	"time"
+)
+
+func TestNodeSorting(t *testing.T) {
+	var now = time.Unix(10000, 0)
+	var zero = *resource.NewQuantity(0, resource.BinarySI)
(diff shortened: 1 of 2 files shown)


j-baker:
Apologies for circle not running - I don't have write access to the repo.

j-baker:
I acquired repo access, closing this out!

PR #77, opened 2019-10-19
Prioritize hosts by current usage rather than just lifetime

j-baker:
We kill our hosts regularly, and so we would like to schedule drivers so
that they heuristically get the longest remaining lifetime. For this
reason, we currently sort our hosts purely in terms of their remaining
host lifetime.
In practice, this is demonstrably wasteful. Consider the case where we
have hosts A and B, where A has a single pod running on it and we've
just booted up host B and so it has no pods running on it. An arriving
job will be scheduled onto host B, leading to poor usage. We would
prefer to schedule the arriving job onto A as well, and be able to kill
B.

This PR takes current usage into account. It fills up the most
utilized hosts first, getting to the near idle ones last.

The goal is that instead of having an even spread of partially used
hosts, we'd be able to partition into totally used hosts and mostly
unused hosts.

Also as part of this, we also change the default to tightly pack executors.
This is our internal default.

Main potential negative: The goal of this is to leave more empty hosts around.
When this happens, we strictly hope to scale down. This means that a given
arriving job will be more likely to run out of space in the cluster, and so there's
potentially more job queuing as a consequence of this. And so I'd imagine that
we'd want to roll this out on a single internal impl first to see how it goes slash
if we need to make any changes in our scaler.

### DIFF
--- a/internal/extender/binpack.go
+++ b/internal/extender/binpack.go
@@ -38,7 +38,7 @@ var binpackFunctions = map[string]*Binpacker{
 func SelectBinpacker(name string) *Binpacker {
 	binpacker, ok := binpackFunctions[name]
 	if !ok {
-		return binpackFunctions[distributeEvenly]
+		return binpackFunctions[tightlyPack]
 	}
 	return binpacker
 }
--- a/internal/extender/binpack.go
+++ b/internal/extender/binpack.go
@@ -38,7 +38,7 @@ var binpackFunctions = map[string]*Binpacker{
 func SelectBinpacker(name string) *Binpacker {
 	binpacker, ok := binpackFunctions[name]
 	if !ok {
-		return binpackFunctions[tightlyPack]
+		return binpackFunctions[distributeEvenly]
 	}
 	return binpacker
 }
--- a/internal/extender/hostordering_test.go
+++ b/internal/extender/hostordering_test.go
@@ -15,32 +15,31 @@
 package extender
 
 import (
-	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
 	"testing"
 
+	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 func TestNodeSorting(t *testing.T) {
-	var zero = *resource.NewQuantity(0, resource.BinarySI)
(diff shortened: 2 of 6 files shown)


internal/extender/resource.go
```diff
@@ -294,11 +292,73 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
...
+	return left.creationTime.Before(right.creationTime)
+}
+
+type scheduleContext struct {
+	creationTime    time.Time
+	availableMemory resource.Quantity
+	availableCPU    resource.Quantity
+}
+
+func cpu(ctx context.Context, r *resources.Resources) resource.Quantity {
```
j-baker: these methods are a bit suspect. Can I guarantee that the result of s.usedResources definitely contains all of the list of nodes I've been handed?

internal/extender/hostordering_test.go
```diff
@@ -0,0 +1,47 @@
...
+		one,
+		one,
+	}
+	var freeMemory = scheduleContext{
+		two,
+		zero,
+	}
+
+	if compareNodes(freeMemory, node) || !compareNodes(node, freeMemory) {
+		t.Error("Young nodes should be sorted by how much memory is available ascending")
```
ashrayjain: change this to be just "nodes" as opposed to "Young nodes"?

internal/extender/hostordering_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	var freeCPU = scheduleContext{
+		one,
+		two,
+	}
+	if compareNodes(freeCPU, node) || !compareNodes(node, freeCPU) {
+		t.Error("If used memory is equal, young nodes should be sorted by how much CPU is available ascending")
```
ashrayjain: same here.

internal/extender/binpack.go
```diff
@@ -38,7 +38,7 @@ var binpackFunctions = map[string]*Binpacker{
 func SelectBinpacker(name string) *Binpacker {
 	binpacker, ok := binpackFunctions[name]
 	if !ok {
-		return binpackFunctions[distributeEvenly]
+		return binpackFunctions[tightlyPack]
```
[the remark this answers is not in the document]
j-baker: fine

PR #81, opened 2019-10-24
Only considers non-schedulable overhead when judging if pod exceeds cluster capacity

rkaram:
We sometimes incorrectly mark a pod as 'exceeding capacity' even though it should have been able to fit in the cluster. The reason for this is a fluctuation in the overhead calculation which appears to be due to the non-deterministic order with which resource reservations and executor pods are deleted.
Both the resource reservation object and the created executors have the driver as an owner. During garbage collection, the resource reservation object can get deleted while the executors stay around, which means that scheduler will count those as pods without reservations, i.e. overhead.

This PR only considers pods that do not have the `spark-scheduler` scheduler name when judging whether new pods would fit in an empty cluster.

### DIFF
--- a/internal/extender/overhead.go
+++ b/internal/extender/overhead.go
@@ -16,13 +16,13 @@ package extender
 
 import (
 	"context"
-	"github.com/palantir/witchcraft-go-error"
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
 	"github.com/palantir/k8s-spark-scheduler/internal/cache"
+	werror "github.com/palantir/witchcraft-go-error"
 	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
 	"github.com/palantir/witchcraft-go-logging/wlog/wapp"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


internal/extender/overhead.go
```diff
@@ -120,32 +135,24 @@ func (o *OverheadComputer) compute(ctx context.Context) {
 			// pending pod or pod succeeded or failed
 			continue
 		}
-		node, err := o.nodeLister.Get(p.Spec.NodeName)
-		if err != nil {
-			svc1log.FromContext(ctx).Warn("node does not exist in cache", svc1log.SafeParam("nodeName", p.Spec.NodeName))
+
+		// found pod with no associated resource reservation, add to overhead
```
laflechejonathan: nit - make the placement of comments on conditional blocks consistents, right now you have both:

```golang
// found a thing that does foo
if thing == foo {
    continue
}

if thing == bar {
    // found a bar thing
}
```

internal/extender/overhead.go
```diff
@@ -120,32 +135,24 @@ func (o *OverheadComputer) compute(ctx context.Context) {
...
+		if err := o.addPodResourcesToGroupResources(ctx, rawOverhead, p); err != nil {
+			svc1log.FromContext(ctx).Warn("could not add pod resources to overhead, skipping", svc1log.SafeParam("failedPod", p.Name), svc1log.Stacktrace(err))
 			continue
```
laflechejonathan: I think it would be a clearer factoring if you did the error-prone thing (getting the node for the pod) separately, so like:

```golang
node, err = getNodeForPod(p);
if err {
    log.Warn("could not get node for pod, skipping")
}

// code to add pod to overhead should not have error handling
```

rkaram: Addressed 👍

internal/extender/overhead.go
```diff
@@ -176,18 +219,24 @@ func podToResources(ctx context.Context, pod *v1.Pod) *resources.Resources {
...
+func (o *OverheadComputer) getPodNodeInstanceGroup(pod *v1.Pod) (string, error) {
+	node, err := o.nodeLister.Get(pod.Spec.NodeName)
+	if err != nil {
+		return "", werror.Wrap(err, "node does not exist in cache", werror.SafeParam("nodeName", pod.Spec.NodeName))
+	}
+	return node.Labels[o.instanceGroupLabel], nil
+}
+
+func (o OverheadComputer) getOverheadByNode(ctx context.Context, overhead Overhead, nodes []*v1.Node) resources.NodeGroupResources {
 	o.overheadLock.RLock()
```
onursatici: I would prefer using the lock at call site, it will make explicit that `overheadLock` guards the overhead.

PR #87, opened 2019-11-06
Prioritize hosts in AZ aware order to reduce cross AZ traffic

a-k-g:
At the moment the allocation of executors for a spark application doesn't take the AZ of the hosts into account at all.

This PR changes the (experimental) sorting to first sort by AZ. We want to avoid distributing the drivers and executors across AZ. We hope this should reduce cross AZ traffic.

Its feature flagged so if it breaks things we can easily roll back.

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -51,6 +51,10 @@ const (
 	// leaderElectionInterval is the default LeaseDuration for core clients.
 	// obtained from k8s.io/component-base/config/v1alpha1
 	leaderElectionInterval = 15 * time.Second
+	// Used to identify an AWS node's zone, example "us-east-1c"
+	// https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#failure-domainbetakubernetesiozone
+	kubernetesFailureDomainLabel = "failure-domain.beta.kubernetes.io/zone"
+	azPlaceholder                = "default"
 )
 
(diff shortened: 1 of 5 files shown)


internal/extender/resource.go
```diff
@@ -311,22 +315,45 @@ func (s *SparkSchedulerExtender) sortNodes(nodes []*v1.Node, availableResources
 		return
 	}
 
-	var nodeNames = make([]string, len(nodes))
+	var nodesByAZ = make(map[string][]*v1.Node)
```
j-baker: can this be a function?

internal/extender/resource.go
```diff
@@ -311,22 +315,45 @@ func (s *SparkSchedulerExtender) sortNodes(nodes []*v1.Node, availableResources
...
 	for i, node := range nodes {
-		nodeNames[i] = node.Name
+		azLabel, ok := node.Labels[kubernetesFailureDomainLabel]
+		if !ok {
+			azLabel = azPlaceholder
+		}
+		nodesByAZ[azLabel] = append(nodesByAZ[azLabel], nodes[i])
+	}
+
+	var allAZLabels = make([]string, 0, len(nodesByAZ))
```
j-baker: likewise, maybe worth having a function that does this? `keys()`?

a-k-g:
Closing in favor of #88

PR #88, opened 2019-11-06
Prioritize hosts in AZ aware order to reduce cross AZ traffic

a-k-g:
At the moment the allocation of executors for a spark application doesn't take the AZ of the hosts into account at all.

This PR changes the (experimental) sorting to first sort by AZ. We want to avoid distributing the drivers and executors across AZ. We hope this should reduce cross AZ traffic.

Its feature flagged so if it breaks things we can easily roll back.

(same as #87)


internal/extender/hostordering_test.go
```diff
@@ -18,30 +18,102 @@ import (
...
+		nil,
+		nil,
+		nil,
+		nil,
+		true,
+		nil,
+		nil,
+		"something",
+		true)
+	mockExtender.sortNodes(actualNodes, availableResources)
```
onursatici: I would make `sortNodes` to be a function instead of a method

internal/extender/resource.go
```diff
@@ -294,15 +298,35 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
...
+// Sort by available resources ascending, with RAM usage more important.
+func lessThan(left *resources.Resources, right *resources.Resources) bool {
 	var memoryCompared = left.Memory.Cmp(right.Memory)
 	if memoryCompared != 0 {
 		return memoryCompared == -1
 	}
 	return left.CPU.Cmp(right.CPU) == -1
 }
 
+func groupNodesByAZ(nodes []*v1.Node) (nodesByAZ map[string][]*v1.Node) {
```
onursatici: now we have non trivial logic for sorting, can we move all sorting related code to its own package?

onursatici: you can also remove the named return value and declare it inside the function body

internal/extender/resource.go
```diff
@@ -294,15 +298,35 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
...
+		azLabel, ok := node.Labels[kubernetesFailureDomainLabel]
+		if !ok {
+			azLabel = azPlaceholder
+		}
+		nodesByAZ[azLabel] = append(nodesByAZ[azLabel], nodes[i])
+	}
+	return nodesByAZ
+}
+
+func getAllAZLabels(nodeGroupsByAZ map[string][]*v1.Node) (azLabels []string) {
```
onursatici: remove the named return value, same as above

internal/extender/nodesorting_test.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+}
+
+func getNodeWithZoneLabel(name, zone string) v1.Node {
+	return v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				"failure-domain.beta.kubernetes.io/zone": zone,
```
ashrayjain: there is a k8s constant for this that we should have used

onursatici: 👍 happy to merge a PR to use the constant instead

PR #98, opened 2019-11-27
Upgrade to new version of k8s-spark-scheduler-lib

a-k-g:
A preview of the changes that would we needed after merging https://github.com/palantir/k8s-spark-scheduler-lib/pull/15

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -447,9 +447,9 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
 	availableNodes := s.getNodes(ctx, nodeNames)
 	usages := s.usedResources()
 	usages.Add(s.overheadComputer.GetOverhead(ctx, availableNodes))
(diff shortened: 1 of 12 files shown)


internal/extender/resource.go
```diff
@@ -445,9 +447,9 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
 	availableNodes := s.getNodes(ctx, nodeNames)
 	usages := s.usedResources()
 	usages.Add(s.overheadComputer.GetOverhead(ctx, availableNodes))
-	availableResources := resources.AvailableForNodes(availableNodes, usages)
+	availableNodesSchedulingMetadata := resources.NodeSchedulingMetadataForNodes(availableNodes, usages)
```
onursatici: I think for rescheduling you can keep the available resources, as nothing else in the metadata is used

onursatici:
Also we should add the new binpacking algo to `binpack.go`

PR #102, opened 2020-01-16
Use node affinity instead of node selector labels from Pod Spec for scheduling

codekarma:
In this change, we switch to using node affinity from PodSpec to read the instance group instead of node selector. We are also now using PodMatchesNodeSelectorAndAffinityTerms to find available nodes.


codekarma:
It is now ready for code review.

internal/podspec.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// MatchPodInstanceGroup compares instance group label on given Pod specs.
+func MatchPodInstanceGroup(pod1 *v1.Pod, pod2 *v1.Pod, instanceGroupLabel string) bool {
+	instanceGroup1, success1 := FindInstanceGroupFromPodSpec(pod1.Spec, instanceGroupLabel)
+	instanceGroup2, success2 := FindInstanceGroupFromPodSpec(pod2.Spec, instanceGroupLabel)
+	return success1 && success1 == success2 && instanceGroup1 == instanceGroup2
```
onursatici: would it be clearer to do `success1 && success2 && instanceGroup1 == instanceGroup2`?

Gopkg.toml
```diff
@@ -18,16 +18,20 @@ required = ["k8s.io/code-generator/cmd/client-gen"]
 
 [[constraint]]
   name = "github.com/palantir/k8s-spark-scheduler-lib"
-  version = "0.2.1"
+  version = "0.2.4"
 
 [[constraint]]
-  version = "kubernetes-1.14.9"
+  version = "kubernetes-1.17.2"
```
onursatici: is it possible to keep this in sync with the k8s dependency version of the lib? which is `0.15.7`: https://github.com/palantir/k8s-spark-scheduler-lib/blob/master/go.mod#L22

internal/extender/resource.go
```diff
@@ -106,15 +107,20 @@ func NewExtender(
...
-	params := internal.PodSafeParams(args.Pod)
+	params := internal.PodSafeParams(*args.Pod)
 	role := args.Pod.Labels[SparkRoleLabel]
-	instanceGroup := args.Pod.Spec.NodeSelector[s.instanceGroupLabel]
-	params["podSparkRole"] = role
-	params["instanceGroup"] = instanceGroup
 	ctx = svc1log.WithLoggerParams(ctx, svc1log.SafeParams(params))
 	logger := svc1log.FromContext(ctx)
+	instanceGroup, success := internal.FindInstanceGroupFromPodSpec(args.Pod.Spec, s.instanceGroupLabel)
+	if !success {
```
onursatici: I think we should preserve the previous behaviour that if the instance group node affinity doesn't exist, we treat it as `""` instead of failing

internal/podspec.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// FindInstanceGroupFromPodSpec extracts the instance group from a Pod spec.
+func FindInstanceGroupFromPodSpec(podSpec v1.PodSpec, instanceGroupLabel string) (instanceGroup string, success bool) {
```
onursatici: I think this should fall back to look for node selectors if no node affinities match the instance group, this will keep backcompat

Gopkg.toml
codekarma: Wonder if it is better to bump lib dependency instead of downgrading here? I can try it out but it took some time finally arriving at this version that works with all dependencies.

internal/podspec.go
codekarma: Yes. This is way better.

internal/extender/resource.go
codekarma: done

internal/podspec.go
codekarma: Make sense.

Gopkg.toml
codekarma: Also found the hard way that all "k8s.io/client-go", "k8s.io/apiextensions-apiserver" and "k8s.io/apimachinery" need to be exact same versions.

onursatici: yea also I am confused on why lib's dependency is shown as `v0.15.7` instead of what it seems to be pulling `v1.15.7`, will have a look.
Alright `1.17.0` means we will support min 1.15 clusters, should be ok

internal/podspec.go
```diff
@@ -0,0 +1,41 @@
...
+
+// MatchPodInstanceGroup compares instance group label on given Pod specs.
+func MatchPodInstanceGroup(pod1 *v1.Pod, pod2 *v1.Pod, instanceGroupLabel string) bool {
+	instanceGroup1, success1 := FindInstanceGroupFromPodSpec(pod1.Spec, instanceGroupLabel)
+	instanceGroup2, success2 := FindInstanceGroupFromPodSpec(pod2.Spec, instanceGroupLabel)
+	return success1 && success2 && instanceGroup1 == instanceGroup2
+}
+
+// FindInstanceGroupFromPodSpec extracts the instance group from a Pod spec.
+func FindInstanceGroupFromPodSpec(podSpec v1.PodSpec, instanceGroupLabel string) (instanceGroup string, success bool) {
```
onursatici: can you remove the named return types as they are not used now?

internal/metrics/queue.go
```diff
@@ -128,7 +129,8 @@ func (p PodHistograms) Inc(key PodTags) {
 
 // MarkTimes inspects pod conditions and marks lifecycle transition times
 func (p PodHistograms) MarkTimes(ctx context.Context, pod *v1.Pod, instanceGroupTagLabel string, now time.Time) {
-	instanceGroupTag := InstanceGroupTag(ctx, pod.Spec.NodeSelector[instanceGroupTagLabel])
+	ig, _ := internal.FindInstanceGroupFromPodSpec(pod.Spec, instanceGroupTagLabel)
```
onursatici: lets do what we are doing in `resource.go` and check err, and set to `""` explicitly

Gopkg.toml
onursatici: this explains the weird k8s versioning: https://github.com/kubernetes/apimachinery/releases/v0.15.7

internal/podspec.go
codekarma: done.

internal/metrics/queue.go
codekarma: done.

PR #104, opened 2020-02-13
Bump witchcraft-go-server to pickup server.uptime metric

blakehawkins:


onursatici:
weird this breaks conjure, I can take this over if you want @blakehawkins

blakehawkins:
Hey @onursatici sorry for the delay here - this should be good to go now

PR #106, opened 2020-03-11
Request retry handling for cache's async client

rkaram:
Adds better request retry handling for cache's async client


internal/cache/async.go
```diff
@@ -103,49 +103,41 @@ func (ac *asyncClient) doUpdate(ctx context.Context, key store.Key) {
...
-		newObj, getErr := ac.client.Get(key.Namespace, key.Name)
+		newObj, getErr := ac.client.Get(r.Key.Namespace, r.Key.Name)
 		switch {
 		case getErr == nil:
 			ac.objectStore.OverrideResourceVersionIfNewer(newObj)
-			ac.doUpdate(ctx, key)
-		case isRetryableError(getErr):
-			svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(getErr))
-			ac.queue.AddIfAbsent(store.UpdateRequest(obj))
+			ac.doUpdate(ctx, store.UpdateRequest(newObj))
```
rkaram: Note that this resets the retry count, we can change that if this is not the behaviour we want.

config/config.go
```diff
@@ -22,13 +22,25 @@ import (
...
+	BinpackAlgo                   string            `yaml:"binpack,omitempty"`
+	InstanceGroupLabel            string            `yaml:"instance-group-label,omitempty"`
+	AsyncClientConfig             AsyncClientConfig `yaml:"async-client-config,omitempty"`
+	UseExperimentalHostPriorities bool              `yaml:"use-experimental-host-priorities,omitempty"`
 
 	ResourceReservationCRDAnnotations map[string]string `yaml:"resource-reservation-crd-annotations,omitempty"`
 }
+
+type AsyncClientConfig struct {
+	maxRetryCount int `yaml:"max-retry-count,omitempty"`
```
onursatici: I think you can use an `*int` here, so you can distinguish not set vs 0 by checking if the value is `nil`, which is the default value for pointers

onursatici: then you can also capitalise the field and access it directly, so you can get rid of the method below

internal/cache/async.go
```diff
@@ -80,20 +82,18 @@ func (ac *asyncClient) doCreate(ctx context.Context, key store.Key) {
...
-		svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
-		ac.queue.AddIfAbsent(store.CreateRequest(obj))
 	case isNamespaceTerminating(err):
 		svc1log.FromContext(ctx).Info("can not create object because its namespace is being terminated")
-		ac.objectStore.Delete(key)
+		ac.objectStore.Delete(r.Key)
 	default:
-		logNonRetryableError(ctx, err)
-		ac.objectStore.Delete(key)
+		if !ac.maybeRetryRequest(ctx, r, err) {
```
onursatici: nit: would be a bit easier to read if you extract the conditional to a variable named `didRetry`, then the fact that 
we are calling `maybeRetryRequest` for the side effect would be clearer.

internal/cache/async.go
```diff
@@ -103,49 +103,41 @@ func (ac *asyncClient) doUpdate(ctx context.Context, key store.Key) {
...
-func isRetryableError(err error) bool {
-	return errors.IsServerTimeout(err) || errors.IsServiceUnavailable(err) ||
-		errors.IsTooManyRequests(err) || errors.IsTimeout(err)
-}
-
-func logNonRetryableError(ctx context.Context, err error) {
-	svc1log.FromContext(ctx).Error("got non retryable error", svc1log.Stacktrace(err))
+func (ac *asyncClient) maybeRetryRequest(ctx context.Context, r store.Request, err error) bool {
+	if r.RetryCount >= ac.config.MaxRetryCount() {
+		svc1log.FromContext(ctx).Error("max retry count reached, dropping request", svc1log.Stacktrace(err))
```
laflechejonathan: We should add metrics on these retry failures, as well as retry attempts

internal/cache/async.go
```diff
@@ -103,49 +103,41 @@ func (ac *asyncClient) doUpdate(ctx context.Context, key store.Key) {
...
+		return false
+	}
+	svc1log.FromContext(ctx).Warn("got retryable error, will retry", svc1log.Stacktrace(err))
+	ac.queue.AddIfAbsent(r.WithIncrementedRetryCount())
```
laflechejonathan: why addIfAsbent here? What is the case where it would not be absent in the queue?

config/config.go
rkaram: @onursatici Updated it to be `*int`, but do you not think it's clearer to keep the method that handles defaulting to 5 in config as a form of documentation and centralising the behaviour?
Kept for now, but happy to change if you think otherwise...

internal/cache/async.go
rkaram: A different client of the cache could have added a request to update the object (for example) while we were processing this one.

internal/cache/async.go
```diff
@@ -149,3 +156,31 @@ func isNamespaceTerminating(err error) bool {
...
+)
+
+// AsyncClientMetrics emits metrics on retries and failures of the internal async client calls to the api server
+// TODO: Move this to the metrics package after a refactor of that package to avoid cyclical imports
+type AsyncClientMetrics struct {
+	ObjectTypeTag string
+}
+
+// MarkRequest marks that a request to the api server is being made
+func (acm *AsyncClientMetrics) MarkRequest(ctx context.Context) {
```
onursatici: I think we should also tag with the request type

rkaram: Done

PR #107, opened 2020-03-24
prevent panic in node affinity check

onursatici:


internal/podspec.go
```diff
@@ -27,7 +27,19 @@ func MatchPodInstanceGroup(pod1 *v1.Pod, pod2 *v1.Pod, instanceGroupLabel string
 
 // FindInstanceGroupFromPodSpec extracts the instance group from a Pod spec.
 func FindInstanceGroupFromPodSpec(podSpec v1.PodSpec, instanceGroupLabel string) (string, bool) {
-	for _, nodeSelectorTerm := range podSpec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
+	affinity := podSpec.Affinity
```
rkaram: Looks like from the current logic, we fall back to `podSpec.NodeSelector[instanceGroupLabel]`. Do we not want this if affinity is not set?

onursatici: ah yea good catch, will fix

PR #108, opened 2020-03-27
Introduces DemandGC 

rkaram:
This PR introduces a background pod event handler which deletes any demand we have previously created for a pod when it gets scheduled.
This is supposed to catch any demands that were not deleted by the extender, and we can later on centralise all demand deletions in here.

### DIFF
--- a/internal/extender/demand_gc.go
+++ b/internal/extender/demand_gc.go
@@ -1,0 +1,109 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Copyright (c) 2020 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extender
+
+import (
+	"context"
+
+	"github.com/palantir/k8s-spark-scheduler/internal"
+	"github.com/palantir/k8s-spark-scheduler/internal/cache"
+	"github.com/palantir/k8s-spark-scheduler/internal/events"
+	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
+	v1 "k8s.io/api/core/v1"
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	clientcache "k8s.io/client-go/tools/cache"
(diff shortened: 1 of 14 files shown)


internal/extender/demand_gc.go
```diff
@@ -0,0 +1,109 @@
...
+	return false
+}
+
+func (dgc *DemandGC) deleteDemandIfExists(pod *v1.Pod) {
+	demandName := demandResourceName(pod)
+	if demand, ok := dgc.demandCache.Get(pod.Namespace, demandName); ok {
+		// there is no harm in the demand being deleted elsewhere in between the two calls.
+		dgc.demandCache.Delete(pod.Namespace, demandName)
+		dgc.logger.Info("Removed demand object because pod is now scheduled", svc1log.SafeParams(internal.DemandSafeParams(demandName, pod.Namespace)))
+		events.EmitDemandDeleted(context.Background(), demand, "DemandGC")
```
rkaram: Will the logger built with `context.Background()` properly record the event?

onursatici: I don't think so, I think you should use the witchcraft context like we do to create the `logger` in `softreservations`

internal/extender/demand_gc.go
```diff
@@ -0,0 +1,109 @@
...
+
+// DemandGC is a background pod event handler which deletes any demand we have previously created for a pod when a pod gets scheduled.
+// We also delete demands elsewhere in the extender when we schedule the pod, but those can miss some demands due to race conditions.
+type DemandGC struct {
+	demandCache *cache.SafeDemandCache
+	ctx         context.Context
+}
+
+// NewDemandGC initializes the DemandGC which handles events in the background
+func NewDemandGC(ctx context.Context, podInformer coreinformers.PodInformer, demandCache *cache.SafeDemandCache) {
```
onursatici: It feels a bit weird to have a function called `NewX` but not return an instance of `X`. I think we should have a `NewDemandGC` function that gets a context and a demand cache, and possibly a method called `startGC` or something like that that accepts an informer

internal/extender/demand_gc.go
```diff
@@ -0,0 +1,109 @@
...
+		clientcache.FilteringResourceEventHandler{
+			FilterFunc: func(obj interface{}) bool {
+				if pod, ok := obj.(*v1.Pod); ok {
+					_, labelFound := pod.Labels[SparkRoleLabel]
+					if labelFound && pod.Spec.SchedulerName == SparkSchedulerName {
+						return true
+					}
+				}
+				return false
+			},
```
onursatici: lets factor this out and get `softReservationStore` to use it as well

internal/extender/demand_gc.go
```diff
@@ -0,0 +1,109 @@
...
+		svc1log.FromContext(dgc.ctx).Error("failed to parse newObj as pod")
+		return
+	}
+
+	if !dgc.isPodScheduled(oldPod) && dgc.isPodScheduled(newPod) {
+		dgc.deleteDemandIfExists(newPod)
+	}
+}
+
+func (dgc *DemandGC) isPodScheduled(pod *v1.Pod) bool {
```
onursatici: I believe you also need the `PodScheduled` condition have a `True` Status, as pods that fail to be scheduled will also have a valid `PodScheduled` condition

internal/extender/demand_gc.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == v1.PodScheduled {
+			return true
+		}
+	}
+	return false
+}
+
+func (dgc *DemandGC) deleteDemandIfExists(pod *v1.Pod) {
```
onursatici: I think we should factor this method out as well, so both the demand GC, extender and the reconciler can call it. It should probably accept a demand cache and a pod. While doing that it would be good to emit the demand deleted event from the reconciler as well

internal/extender/demand_gc.go
rkaram: Good point, there is no strong reason to have a method which returns an instance of the object since the only current use case is starting the background GC. Renamed this to `StartDemandGC`

PR #116, opened 2020-05-03
Refactors Reservation creations into a ResourceReservationManager

rkaram:
This PR refactors the way we get and create both resource reservations as well as soft reservations. The aim of this is two fold:
1. Simplify the logic in the scheduling code paths by abstracting away common functionality such as checking for unbound reservations, checking for free extra executor space, patching the reservation object, etc...
2. Introduce a common place through which all reads and writes of reservations go through which will enable us to introduce new functionality in the future

This PR does not currently include refactoring the failover code, but we should handle this in a separate PR.

### DIFF
--- a/internal/common/utils/sets.go
+++ b/internal/common/utils/sets.go
@@ -49,7 +49,7 @@ func (s *StringSet) Size() int {
 // ToSlice returns all the elements of StringSet as a slice of strings
 func (s *StringSet) ToSlice() []string {
 	result := make([]string, len(s.elements))
-	for e, _ := range s.elements {
+	for e := range s.elements {
 		result = append(result, e)
 	}
 	return result
--- a/internal/common/utils/sets.go
+++ b/internal/common/utils/sets.go
@@ -1,0 +1,56 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+// StringSet is a non-thread safe set of unique strings
+type StringSet struct {
+	elements map[string]bool
+}
+
+// NewStringSet constructs and returns a StringSet
+func NewStringSet(size int) *StringSet {
+	return &StringSet{
+		elements: make(map[string]bool, size),
+	}
+}
+
+// Add adds the string e to the StringSet if it is not already there
+func (s *StringSet) Add(e string) {
+	s.elements[e] = true
+}
+
+// Remove removes the string e from the StringSet if it is there. It is a no-op otherwise
+func (s *StringSet) Remove(e string) {
+	delete(s.elements, e)
+}
+
+// Contains returns true if the string e is in the StringSet, false otherwise
+func (s *StringSet) Contains(e string) bool {
+	return s.elements[e]
+}
+
+// Size returns the number of elements in the StringSet
+func (s *StringSet) Size() int {
+	return len(s.elements)
+}
+
+// ToSlice returns all the elements of StringSet as a slice of strings
+func (s *StringSet) ToSlice() []string {
+	result := make([]string, len(s.elements))
+	for e, _ := range s.elements {
+		result = append(result, e)
+	}
+	return result
+}
--- a/internal/common/utils/sets.go
+++ b/internal/common/utils/sets.go
@@ -15,41 +15,44 @@
 package utils
 
 // StringSet is a non-thread safe set of unique strings
-type StringSet struct {
-	elements map[string]bool
-}
+type StringSet map[string]bool
 
 // NewStringSet constructs and returns a StringSet
-func NewStringSet(size int) *StringSet {
-	return &StringSet{
-		elements: make(map[string]bool, size),
-	}
+func NewStringSet(size int) StringSet {
+	return make(StringSet, size)
 }
 
 // Add adds the string e to the StringSet if it is not already there
-func (s *StringSet) Add(e string) {
-	s.elements[e] = true
+func (s StringSet) Add(e string) {
+	s[e] = true
+}
+
+// AddAll adds the strings in es to the StringSet if they are not already there
+func (s StringSet) AddAll(es []string) {
+	for _, e := range es {
+		s[e] = true
+	}
 }
 
 // Remove removes the string e from the StringSet if it is there. It is a no-op otherwise
-func (s *StringSet) Remove(e string) {
-	delete(s.elements, e)
+func (s StringSet) Remove(e string) {
+	delete(s, e)
 }
 
 // Contains returns true if the string e is in the StringSet, false otherwise
-func (s *StringSet) Contains(e string) bool {
-	return s.elements[e]
+func (s StringSet) Contains(e string) bool {
+	return s[e]
 }
 
 // Size returns the number of elements in the StringSet
-func (s *StringSet) Size() int {
-	return len(s.elements)
+func (s StringSet) Size() int {
+	return len(s)
 }
 
 // ToSlice returns all the elements of StringSet as a slice of strings
-func (s *StringSet) ToSlice() []string {
-	result := make([]string, len(s.elements))
-	for e := range s.elements {
+func (s StringSet) ToSlice() []string {
+	result := make([]string, s.Size())
+	for e := range s {
 		result = append(result, e)
 	}
 	return result
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -369,7 +369,7 @@ func (s *SparkSchedulerExtender) selectExecutorNode(ctx context.Context, executo
 		return "", failureInternal, werror.WrapWithContextParams(ctx, err, "error when checking for free executor spots")
 	}
 	if freeExecutorSpots > 0 {
-		isExtraExecutor := len(unboundReservationNodes) > 0			// Dynamic allocation executor since there were no free unbound reservations
+		isExtraExecutor := len(unboundReservationNodes) > 0 // Dynamic allocation executor since there were no free unbound reservations
 		nodeName, outcome, err := s.rescheduleExecutor(ctx, executor, nodeNames, isExtraExecutor)
 		if err != nil {
 			return "", outcome, werror.WrapWithContextParams(ctx, err, "failed to reschedule executor")
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -361,7 +361,7 @@ func (s *SparkSchedulerExtender) selectExecutorNode(ctx context.Context, executo
 	if foundUnbound {
 		if resultNode, ok := s.getReservationNodeFromNodeList(unboundReservationNodes, nodeNames); ok {
 			svc1log.FromContext(ctx).Info("found unbound reservation node for executor", svc1log.SafeParam("nodeName", resultNode))
-			err := s.resourceReservationManager.ReserveForExecutor(ctx, executor, resultNode)
+			err := s.resourceReservationManager.ReserveForExecutorOnUnboundReservation(ctx, executor, resultNode)
 			if err != nil {
 				return "", failureInternal, werror.WrapWithContextParams(ctx, err, "failed to reserve node for executor")
 			}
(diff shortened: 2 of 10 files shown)


internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	for _, pod := range pods {
+		if !rrm.IsPodTerminated(pod) {
+			activePodNames[pod.Name] = true
+		}
+	}
+	return activePodNames, nil
+}
+
+// IsPodTerminated returns whether the pod is considered to be terminated
+func (rrm *ResourceReservationManager) IsPodTerminated(pod *v1.Pod) bool {
```
rkaram, line 275: The following three methods (`IsPodTerminated`, `NewResourceReservation` and `ExecutorReservationName`) should not be exported. Unfortunately, they're being used in failover and refactoring failover to use `ResourceReservationManager` is out of the scope of this PR. 
We should consider limiting them back as soon as we refactor failover.

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
-		copyResourceReservation.Spec.Reservations[unboundReservation] = reservation
+		s.removeDemandIfExists(ctx, executor)
+		return nodeName, successRescheduled, nil
 	}
 
-	copyResourceReservation.Status.Pods[unboundReservation] = executor.Name
-	err := s.resourceReservations.Update(copyResourceReservation)
-	if err != nil {
-		return "", failureInternal, werror.Wrap(err, "failed to update resource reservation")
+	return "", failureFit, werror.ErrorWithContextParams(ctx, "application has no free executor spots to schedule this one")
```
rkaram, line 385: We might want to pick a different outcome here rather than `failureFit`. I don't believe we previously had an equivalent to this as we always seem to reschedule.

rkaram: Changed to `failureUnbound` which seems to have been the one we used in those cases.

internal/common/utils/sets.go
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+// StringSet is a non-thread safe set of unique strings
+type StringSet struct {
```
onursatici, line 18: It might be a bit clearer if you define this as `type StringSet map[string]bool`, that way you can get rid of the struct and don't have a pointer to a map

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
-	unboundReservations, outcome, unboundResErr := s.findUnboundReservations(ctx, executor, resourceReservation)
-	if unboundResErr != nil {
-		extraExecutorCount := 0
-		if sr, ok := s.softReservationStore.GetSoftReservation(executor.Labels[common.SparkAppIDLabel]); ok {
-			extraExecutorCount = len(sr.Reservations)
-		}
-		driver, err := s.podLister.getDriverPod(ctx, executor)
-		if err != nil {
-			return "", failureInternal, err
+	if alreadyBoundNode != "" {
```
onursatici, line 344: should we signal this with another return value from `FindAlreadyBoundReservationNode`, like it can return the `nodeName, found, error` and if `found` we can rely on a valid non empty `nodeName`

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	return len(unboundReservations) + extraExecutorFreeSpots, nil
+}
+
+// ReserveForExecutor creates a reservation for the passed executor on the passed node. This reservation could either be a resource reservation,
+// or a soft reservation if dynamic allocation is enabled.
+func (rrm *ResourceReservationManager) ReserveForExecutor(ctx context.Context, executor *v1.Pod, node string) error {
+	rrm.mutex.Lock()
+	defer rrm.mutex.Unlock()
+
+	// TODO: make sure executor doesn't already have a reservation, if it does, it must be freed (we don't seem to handle this in the current logic, but we should going forward)
```
onursatici, line 145: don't we handle this at `FindAlreadyBoundReservationNode`?

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	resourceReservations := rrm.resourceReservations.List()
+	usage := resources.UsageForNodes(resourceReservations)
+	usage.Add(rrm.softReservationStore.UsedSoftReservationResources())
+	return usage
+}
+
+func (rrm *ResourceReservationManager) bindExecutorToResourceReservation(ctx context.Context, executor *v1.Pod, reservationName string, node string) error {
+	resourceReservation, ok := rrm.GetResourceReservation(executor)
+	if !ok {
+		return werror.ErrorWithContextParams(ctx, "failed to get resource reservationName")
```
onursatici, line 188: pass reservationName as a param

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	}
+	copyResourceReservation := resourceReservation.DeepCopy()
+	reservationObject := copyResourceReservation.Spec.Reservations[reservationName]
+	reservationObject.Node = node
+	copyResourceReservation.Spec.Reservations[reservationName] = reservationObject
+	copyResourceReservation.Status.Pods[reservationName] = executor.Name
+	err := rrm.resourceReservations.Update(copyResourceReservation)
+	if err != nil {
+		return werror.WrapWithContextParams(ctx, err, "failed to update resource reservationName")
```
onursatici, line 197: pass reservationName as a param

onursatici: actually we can probably not wrap and directly return the result of the `Update` call, as we also wrap the error at `resource.go`

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	unboundReservationsToNodes, err := rrm.getUnboundReservations(ctx, executor)
+	if err != nil {
+		return err
+	}
+	// Sort the unbound reservations such that we favor reservations that are already tied to the requested node first
+	sortedUnboundReservations := make([]string, len(unboundReservationsToNodes))
+	for reservation := range unboundReservationsToNodes {
+		sortedUnboundReservations = append(sortedUnboundReservations, reservation)
+	}
+	sort.Slice(sortedUnboundReservations, func(i, j int) bool {
```
onursatici, line 156: does this mean we can bind this executor to some other node than `node`? Wouldn't that break the guarantee that the extender will always return a node that exists in the list of given candidate nodes?

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
+		// check that it is part of the nodes passed and return it
+		if resultNode, ok := s.getReservationNodeFromNodeList([]string{alreadyBoundNode}, nodeNames); ok {
```
laflechejonathan, line 346: in the case where this condition is false, what happens?

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
 	}
-	var unboundReservation string
-	for _, name := range nodeNames {
-		if reservation, ok := nodeToReservation[name]; ok {
-			unboundReservation = reservation
-			break
-		}
+
+	// Else, check if you still can have an executor, and if yes, reschedule
+	freeExecutorSpots, err := s.resourceReservationManager.GetFreeExecutorSpots(ctx, executor)
```
laflechejonathan, line 367: What's the difference between `freeExecutorSpots` and `unboundReservations`?

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
-		// try to reschedule the executor, breaking FIFO, but preventing executor starvation
-		// we are guaranteed len(unboundResourceReservations) > 0
-		unboundReservation = unboundReservations[0]
-		driver, err := s.podLister.getDriverPod(ctx, executor)
-		if err != nil {
-			return "", failureInternal, err
-		}
-		sparkResources, err := sparkResources(ctx, driver)
+	if freeExecutorSpots > 0 {
+		isExtraExecutor := len(unboundReservationNodes) > 0 // Dynamic allocation executor since there were no free unbound reservations
```
laflechejonathan, line 372: what's a case where freeExecutorSpots is greater than zero, it's not dynamic allocation, and there are no unbound reservations?

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
+		err = s.resourceReservationManager.ReserveForExecutor(ctx, executor, nodeName)
 		if err != nil {
-			return "", outcome, err
+			return "", failureInternal, werror.WrapWithContextParams(ctx, err, "failed to reserve node for rescheduled executor")
 		}
-		svc1log.FromContext(ctx).Info("rescheduling executor", svc1log.SafeParam("reservationName", unboundReservation), svc1log.SafeParam("node", node))
-		reservation := copyResourceReservation.Spec.Reservations[unboundReservation]
-		reservation.Node = node
-		copyResourceReservation.Spec.Reservations[unboundReservation] = reservation
+		s.removeDemandIfExists(ctx, executor)
```
laflechejonathan, line 381: It seems surprising that this is buried deep here - feels like it should just be an invariant very high up in the code, that if we are about to return a successful schedule, we `removeDemandIfExists`

internal/extender/resource.go
```diff
@@ -336,92 +337,66 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
...
-	copyResourceReservation.Status.Pods[unboundReservation] = executor.Name
-	err := s.resourceReservations.Update(copyResourceReservation)
-	if err != nil {
-		return "", failureInternal, werror.Wrap(err, "failed to update resource reservation")
+	return "", failureUnbound, werror.ErrorWithContextParams(ctx, "application has no free executor spots to schedule this one")
+}
+
+// getReservationNodeFromNodeList filters the list of reservationNodes to return a single one that also appears in nodeNames, or false.
+func (s *SparkSchedulerExtender) getReservationNodeFromNodeList(reservationNodes []string, nodeNames []string) (string, bool) {
+	reservationNodeSet := utils.NewStringSet(len(reservationNodes))
```
laflechejonathan, line 390: nit - should you have a constructor that takes a collection and returns the corresponding string set?

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+	for reservationName, reservation := range resourceReservation.Spec.Reservations {
+		podIdentifier, ok := resourceReservation.Status.Pods[reservationName]
+		if !ok || !activePodNames[podIdentifier] {
+			unboundReservationsToNodes[reservationName] = reservation.Node
+		}
+	}
+	return unboundReservationsToNodes, nil
+}
+
+func (rrm *ResourceReservationManager) getFreeExtraExecutorSpots(ctx context.Context, executor *v1.Pod) (int, error) {
```
laflechejonathan, line 239: I think instead of introducing the concept of `ExtraExecutors` you should use the existing terminology, either `getSoftReservationExecutorSpots` or `getDynamicAllocatedExecutorSpots`

internal/extender/resource.go
rkaram: If this is false, that means that we have a previously bound node for this executor, but we did not get it from `kube-scheduler` as a possible node (because it could no longer be available for example), so we move on and try to find other unbound reservations to take.

internal/extender/resource.go
rkaram: I had done this initially, but then found it a bit redundant, although I agree it is more proper so will do this

internal/extender/resource.go
rkaram: `unboundReservationNodes` is a set of nodes that are reserved but that don't have executors currently scheduled on them.
`GetFreeExecutorSpots` returns the number of remaining executors this Spark application is allowed to schedule. Agreed that the naming is confusing, renamed to `GetRemainingAllowedExecutorCount`.

internal/extender/resource.go
rkaram: Good point. The initial intention was to cover the case where there are unbound reservations, but they were not passed to us as potential nodes, so we would get in a situation where we get here with `freeExecutorSpots > 0` even though there are unbound reservations previously found.
I notice though that we can't consider this a dynamic allocation executor, so will fix this.

internal/extender/resource.go
rkaram: Added an `AddAll(es []string)` method to `StringSet`

internal/extender/resourcereservations.go
```diff
@@ -15,26 +15,281 @@
...
+
+// CreateReservations creates the necessary reservations for an application whether those are resource reservation objects or
+// in-memory soft reservations for extra executors.
+func (rrm *ResourceReservationManager) CreateReservations(
+	ctx context.Context,
+	driver *v1.Pod,
+	applicationResources *sparkApplicationResources,
+	driverNode string,
+	executorNodes []string) (*v1beta1.ResourceReservation, error) {
+	rr, ok := rrm.GetResourceReservation(driver)
```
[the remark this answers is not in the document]
rkaram: We are, but the intent was to have the abstractions in `ResourceReservationManager` be separate and do their own checks before creating the reservation, especially that we could still want to create the soft reservation object even though the resource reservation one exists (in say failover).
Let me know if you think this is expensive.

internal/extender/resourcereservations.go
rkaram: We could find an already bound reservation, but for a node that is no longer available. In this case, we will reschedule the executor onto a new node and call `ReserveForExecutor` for this one.
This means that we will double reserve nodes for a single executor.

internal/extender/resourcereservations.go
rkaram: `ReserveForExecutor` will only bind the executor to the passed node. 
The sorting in this case would prefer choosing the **reservation name** to pick for the (executor, node) pair as a reservation that was previously attached to the same node rather than replacing another node in a separate executor.
Note that I don't think this is important for correctness, I just found that it would be better to do this since the alternative means the nodes in the resource reservation object we first create would frequently be overriden when it is not needed. Happy to remove if you think the added complexity does not bring value.

internal/extender/resourcereservations.go
rkaram: The extra message might be clearer when wrapping.

internal/extender/resourcereservations.go
rkaram: The initial intent was to prepare for moving all reservation creations to go through `ResourceReservationManager` which is why I made those private methods to this. 
We can do this at a later date though when refactoring failover, reverted for now.

internal/extender/resource.go
rkaram: Fixed to check the opposite, i.e. that there are no unbound reservations found

internal/extender/resourcereservations.go
rkaram: Changed this to 2 separate methods each without a sorting to make the logic explicit rather than implicit.

internal/extender/resourcereservations.go
rkaram: Fixed by counting reservations bound to executors on a node different from the reservation node, as an unbound reservation.

PR #117, opened 2020-05-05
Lazy demand informer

onursatici:


internal/crd/demand_informer.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, false
+	}
+}
+
+// Ready returns a channel that will be closed when the informer is initialized
+func (ldi *LazyDemandInformer) Ready() <-chan struct{} {
+	return ldi.ready
+}
+
+// Run starts the goroutine to check for the existence of the demand CRD,
```
laflechejonathan, line 77: `starts the goroutine to check for the existence of the demand CRD`, but seems like it's happenning synchronously?

internal/crd/demand_informer.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			svc1log.FromContext(ctx).Error("failed initializing demand informer", svc1log.Stacktrace(err))
+			return false
+		}
+	}
+	return ready
+}
+
+func (ldi *LazyDemandInformer) initializeInformer(ctx context.Context) error {
+	ldi.lock.Lock()
```
laflechejonathan, line 121: I don't think I understand why this lock is needed, can you explain?

internal/crd/demand_informer.go
```diff
@@ -0,0 +1,142 @@
...
+			return werror.ErrorWithContextParams(ctx, "timeout syncing informer", werror.SafeParam("timeoutSeconds", informerSyncTimeout.Seconds()))
+		}
+		return nil
+	}, retry.WithMaxAttempts(informerSyncRetryCount), retry.WithInitialBackoff(informerSyncRetryInitialBackoff))
+
+	if err != nil {
+		return err
+	}
+	ldi.informer = informerInterface
+	close(ldi.ready)
```
laflechejonathan, line 140: It took me a while to find where this was being done, think you should move it to `doStart` and let this be a pure function that just creates and returns the informer.

internal/crd/demand_informer.go
onursatici: it does the first retry synchronously, if failed starts the goroutine to check periodically

internal/crd/demand_informer.go
onursatici: yea I wanted to block the reader goroutines if there was an ongoing call to this method, but I think it is fine to remove it. Actual guard for the informer field is the ready channel and all readers should wait on the ready channel before reading

internal/crd/demand_informer.go
onursatici: Then I thought we would need to close in two different places, in `doStart` for the async case, and in `Run`, before spawning the goroutine for the synchronous case

internal/cache/safedemands.go
```diff
@@ -16,123 +16,72 @@ package cache
...
-	defer t.Stop()
-	for {
-		select {
-		case <-ctx.Done():
-			return nil
-		case <-t.C:
-			if sdc.checkDemandCRDExists(ctx) {
-				return nil
-			}
+		err := wapp.RunWithFatalLogging(ctx, sdc.wait)
```
laflechejonathan, line 60: I think it would be clear if this read like:

```golang
sdc.initializeCache(ctx);
go func() {
   wapp.RunWithFatalLogging(ctx, sdc.wait);
   sdc.initializeCache(ctx);
}()
```
(ignoring error handling)

PR #121, opened 2020-05-11
Dynamic Allocation Executor Compaction

rkaram:
This PR adds an informer handler for executor deaths and moves soft reservations to take that executor's resource reservations in order to compact reserved space.

**Details**
Dynamic allocation executors are handled by reserving for them in-memory (soft reservations) rather than in the resource reservation CRD. 
When an executor with a resource reservation dies, we still count that reservation + the soft reservations towards the cluster reserved resources. The problem is that this executor may never be replaced by Spark and we would not want to keep reserving for it and overcounting our usage.

### DIFF
--- a/internal/common/utils/pods.go
+++ b/internal/common/utils/pods.go
@@ -54,7 +54,7 @@ func GetPodFromObjectOrTombstone(obj interface{}) (*v1.Pod, bool) {
 }
 
 func getRoleIfSparkSchedulerPod(obj interface{}) (string, bool) {
-	if pod, ok := GetPodFromObjectOrTombstone(obj); ok {
+	if pod, ok := obj.(*v1.Pod); ok {
 		role, labelFound := pod.Labels[common.SparkRoleLabel]
 		if labelFound && pod.Spec.SchedulerName == common.SparkSchedulerName {
 			return role, true
--- a/internal/common/utils/pods.go
+++ b/internal/common/utils/pods.go
@@ -15,7 +15,10 @@
 package utils
 
 import (
+	"context"
+
 	"github.com/palantir/k8s-spark-scheduler/internal/common"
+	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
 	v1 "k8s.io/api/core/v1"
 	clientcache "k8s.io/client-go/tools/cache"
 )
@@ -58,4 +61,41 @@ func getRoleIfSparkSchedulerPod(obj interface{}) (string, bool) {
 		}
 	}
 	return "", false
+}
+
+// IsPodTerminated returns whether the pod is considered to be terminated
+func IsPodTerminated(pod *v1.Pod) bool {
+	allTerminated := len(pod.Status.ContainerStatuses) > 0
+	for _, status := range pod.Status.ContainerStatuses {
+		allTerminated = allTerminated && status.State.Terminated != nil
+	}
+	return allTerminated
+}
+
+// OnPodScheduled returns a function that calls the wrapped function if the pod is scheduled
+func OnPodScheduled(ctx context.Context, fn func(*v1.Pod)) func(interface{}, interface{}) {
+	return func(oldObj interface{}, newObj interface{}) {
+		oldPod, ok := oldObj.(*v1.Pod)
+		if !ok {
+			svc1log.FromContext(ctx).Error("failed to parse oldObj as pod")
+			return
+		}
+		newPod, ok := newObj.(*v1.Pod)
+		if !ok {
+			svc1log.FromContext(ctx).Error("failed to parse newObj as pod")
+			return
+		}
+		if !isPodScheduled(oldPod) && isPodScheduled(newPod) {
+			fn(newPod)
+		}
+	}
+}
+
+func isPodScheduled(pod *v1.Pod) bool {
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == v1.PodScheduled && cond.Status == v1.ConditionTrue {
+			return true
+		}
+	}
+	return false
 }
--- a/internal/common/utils/pods.go
+++ b/internal/common/utils/pods.go
@@ -26,6 +26,7 @@ func IsSparkSchedulerPod(obj interface{}) bool {
 	return isSparkSchedulerPod
 }
 
+// IsSparkSchedulerExecutorPod returns whether the passed object is a spark application pod which has this scheduler in the scheduler spec and is an executor
 func IsSparkSchedulerExecutorPod(obj interface{}) bool {
 	if role, isSparkSchedulerPod := getRoleIfSparkSchedulerPod(obj); isSparkSchedulerPod {
 		return role == common.Executor
@@ -33,6 +34,7 @@ func IsSparkSchedulerExecutorPod(obj interface{}) bool {
 	return false
 }
 
+// GetPodFromObjectOrTombstone tries to cast the passed object to a Pod object, and if that's not possible tries to get the Pod object from the tombstone
 func GetPodFromObjectOrTombstone(obj interface{}) (*v1.Pod, bool) {
 	pod, ok := obj.(*v1.Pod)
 	if !ok {
--- a/internal/common/utils/pods.go
+++ b/internal/common/utils/pods.go
@@ -17,15 +17,43 @@ package utils
 import (
 	"github.com/palantir/k8s-spark-scheduler/internal/common"
 	v1 "k8s.io/api/core/v1"
+	clientcache "k8s.io/client-go/tools/cache"
 )
 
 // IsSparkSchedulerPod returns whether the passed object is a spark application pod which has this scheduler in the scheduler spec
 func IsSparkSchedulerPod(obj interface{}) bool {
-	if pod, ok := obj.(*v1.Pod); ok {
-		_, labelFound := pod.Labels[common.SparkRoleLabel]
-		if labelFound && pod.Spec.SchedulerName == common.SparkSchedulerName {
-			return true
-		}
+	_, isSparkSchedulerPod := getRoleIfSparkSchedulerPod(obj)
+	return isSparkSchedulerPod
+}
+
+func IsSparkSchedulerExecutorPod(obj interface{}) bool {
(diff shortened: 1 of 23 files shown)


internal/common/utils/pods.go
```diff
@@ -20,17 +20,47 @@ import (
...
+		pod, ok = tombstone.Obj.(*v1.Pod)
+		if !ok {
+			return nil, false
+		}
+	}
+	return pod, true
+}
+
+func getRoleIfSparkSchedulerPod(obj interface{}) (string, bool) {
+	if pod, ok := GetPodFromObjectOrTombstone(obj); ok {
```
onursatici, line 57: I think we should just try to cast this to a pod object and not try to extract it from a tombstone.
We use `IsSparkSchedulerPod` and `IsSparkSchedulerExecutorPod` in `FilteringResourceEventHandlers` and now we will return true for dead pods as well, but in order for the `DeleteFunc` to be called on pod deletions, I believe the filter should return false as soon as the pod is deleted

internal/extender/resource.go
```diff
@@ -341,6 +342,7 @@ func (s *SparkSchedulerExtender) potentialNodes(availableNodesSchedulingMetadata
 }
 
 func (s *SparkSchedulerExtender) selectExecutorNode(ctx context.Context, executor *v1.Pod, nodeNames []string) (string, string, error) {
+	appID := executor.Labels[common.SparkAppIDLabel]
```
onursatici, line 345: you can inline this

internal/extender/resourcereservations.go
```diff
@@ -283,6 +386,27 @@ func (rrm *ResourceReservationManager) getActivePods(ctx context.Context, pod *v
 	return activePodNames, nil
 }
 
+func (rrm *ResourceReservationManager) onExecutorPodDeletion(obj interface{}) {
+	pod, ok := utils.GetPodFromObjectOrTombstone(obj)
+	if !ok {
+		rrm.logger.Warn("failed to parse object as pod, skipping")
```
onursatici, line 392: lets keep the only usage of `rrm.logger` here because we don't have a context, but for the rest of the logs, lets use the `ctx` we get from the predicate call

internal/extender/resourcereservations.go
```diff
@@ -283,6 +386,27 @@ func (rrm *ResourceReservationManager) getActivePods(ctx context.Context, pod *v
...
+		return
+	}
+
+	// Only keep executor pods which are for dynamic allocation applications and which have resource reservations to give back
+	if _, ok = rrm.softReservationStore.GetSoftReservation(pod.Labels[common.SparkAppIDLabel]); ok {
+		if !rrm.softReservationStore.ExecutorHasSoftReservation(context.Background(), pod) {
```
onursatici, line 398: we can save the `ctx` instead of the `logger` and use it here, with the background context I believe logs will be dropped

internal/extender/resourcereservations.go
```diff
@@ -283,6 +386,27 @@ func (rrm *ResourceReservationManager) getActivePods(ctx context.Context, pod *v
...
+			rrm.addPodForDynamicAllocationCompaction(pod)
+		}
+	}
+}
+
+func (rrm *ResourceReservationManager) addPodForDynamicAllocationCompaction(pod *v1.Pod) {
+	rrm.dynamicAllocationCompactionSliceLock.Lock()
+	defer rrm.dynamicAllocationCompactionSliceLock.Unlock()
+	rrm.dynamicAllocationCompactionPods = append(rrm.dynamicAllocationCompactionPods, sparkAppIdentifier{pod.Namespace, pod.Labels[common.SparkAppIDLabel]})
```
onursatici, line 407: Because we are tracking at app id level, can we just use a map and drain that instead of a slice? that way you should be able to remove `sparkAppIdentifier` and keep a `map[appId]namespace` instead

internal/extender/resourcereservations.go
```diff
@@ -26,38 +26,63 @@ import (
...
 // ResourceReservationManager is a central point which manages the creation and reading of both resource reservations and soft reservations
 type ResourceReservationManager struct {
-	resourceReservations *cache.ResourceReservationCache
-	softReservationStore *cache.SoftReservationStore
-	podLister            *SparkPodLister
-	mutex                sync.RWMutex
+	resourceReservations                 *cache.ResourceReservationCache
+	softReservationStore                 *cache.SoftReservationStore
+	podLister                            *SparkPodLister
+	mutex                                sync.RWMutex
```
onursatici, line 51: should we make both mutexes `sync.Mutex` as we don't use `RLock` and `RUnlock`?

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+// reservation objects and report correctly on reserved usage.
+func (rrm *ResourceReservationManager) CompactDynamicAllocationApplications(ctx context.Context) {
+	timer := metrics.GetAndStartSoftReservationCompactionTimer()
+	defer timer.MarkCompactionComplete(ctx)
+	dynamicAllocationAppsToCompact := rrm.drainDynamicAllocationCompactionSlice()
+
+	rrm.mutex.Lock()
+	defer rrm.mutex.Unlock()
+	for _, app := range dynamicAllocationAppsToCompact {
+		if sr, ok := rrm.softReservationStore.GetSoftReservation(app.appID); ok {
```
onursatici, line 230: lets invert the conditional, log and continue to reduce nesting

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+			pods, err := rrm.getActivePods(ctx, app.namespace, app.appID)
+			if err != nil {
+				rrm.logger.Error("error getting active pods during compaction", svc1log.SafeParam("podNamespace", app.namespace), svc1log.Stacktrace(err))
+				continue
+			}
+			for podName := range sr.Reservations {
+				pod, ok := pods[podName]
+				if !ok {
+					rrm.logger.Info("executor pod with soft reservation no longer active, skipping compaction for this one",
+						svc1log.SafeParam("podNamespace", pod.Namespace),
```
onursatici, line 241: these will clash with the top level params we add in the `Predicate` function, we should rename them to something like`podNamespaceUnderCompation`

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+					continue
+				}
+				rrm.compactSoftReservationPod(ctx, pod)
+			}
+		}
+	}
+}
+
+func (rrm *ResourceReservationManager) compactSoftReservationPod(ctx context.Context, pod *v1.Pod) {
+	appID := pod.Labels[common.SparkAppIDLabel]
```
onursatici, line 252: you can pass `appID` as an argument as you already know that at call site

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+	if err != nil {
+		rrm.logger.Error("failed to get unbound reservations for executor", svc1log.SafeParam("podNamespace", pod.Namespace),
+			svc1log.SafeParam("podName", pod.Name), svc1log.Stacktrace(err))
+		return
+	}
+	if len(unboundReservationsToNodes) > 0 {
+		for reservationName, reservationNode := range unboundReservationsToNodes {
+			if reservationNode == pod.Spec.NodeName {
+				rrm.logger.Info("compacting executor soft reservation to resource reservation",
+					svc1log.SafeParam("podNamespace", pod.Namespace), svc1log.SafeParam("podName", pod.Name),
```
onursatici, line 263: These param names also need to be more specific to disambiguate

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+					svc1log.SafeParam("nodeName", pod.Spec.NodeName), svc1log.SafeParam("reservationName", reservationName))
+				err := rrm.bindExecutorToResourceReservation(ctx, pod, reservationName, pod.Spec.NodeName)
```
onursatici, line 265: lets pass `reservationNode` here instead of `pod.Spec.NodeName` to make it more clear that we are not rescheduling

internal/extender/resourcereservations.go
```diff
@@ -191,8 +216,86 @@ func (rrm *ResourceReservationManager) GetReservedResources() resources.NodeGrou
...
+						svc1log.SafeParam("podName", podName))
+					continue
+				}
+				rrm.compactSoftReservationPod(ctx, pod)
+			}
+		}
+	}
+}
+
+func (rrm *ResourceReservationManager) compactSoftReservationPod(ctx context.Context, pod *v1.Pod) {
```
onursatici, line 251: can you add a comment saying that this has to be called under `rrm.mutex` locked?

internal/extender/resourcereservations.go
rkaram: Discussed and decided to keep as is since it will not clash given we're using the `ResourceReservationManager` logger which we construct with another context

internal/extender/resourcereservations.go
rkaram: Discussed and decided to keep as is since it will not clash given we're using the `ResourceReservationManager` logger which we construct with another context

internal/extender/resourcereservations.go
rkaram: Discussed and decided to keep as is since it might be less confusing to have the compaction logs be separate from the predicate context and the params added to it. My view is that the compaction is a 'background' process, and the only reason to currently initiate it from the Predicate call is that we cannot be sure we are the leader until we get a request.

internal/extender/resourcereservations.go
rkaram: Yes, but we still need the pod object passed so thought about just extracting it instead of taking additional params. What do you think?

internal/extender/resourcereservations.go
rkaram: Good point, done

PR #130, opened 2020-07-03
Detailed after demand waste metric with failure outcomes

rkaram:
## Before
The current `after-demand-fulfilled` waste metric tracks the time it took a pod to get scheduled after a demand was fulfilled, indicating slowness somewhere in scheduler or our extender. The problem is that this dismisses other factors such as FIFO requirements where the demand might have been fulfilled, but we still can't schedule the pod due to an earlier one that was still not scheduled.

## After
We differentiate between waste after demand fulfilled because of scheduler slowness and other scheduling failure outcomes.

## Notes
This uses a best effort in-memory tracker of last failure outcome since we don't commit the result anywhere and it would not currently be worth publishing object updates for this specific metric.

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -132,8 +132,7 @@ func (s *SparkSchedulerExtender) Predicate(ctx context.Context, args schedulerap
 	if err != nil {
 		msg := "failed to reconcile"
 		logger.Error(msg, svc1log.Stacktrace(err))
-		s.wasteMetricsReporter.MarkFailedSchedulingAttempt(args.Pod, failureInternal)
-		return failWithMessage(ctx, args, msg)
+		return s.failWithMessage(ctx, failureInternal, args, msg)
 	}
 	s.resourceReservationManager.CompactDynamicAllocationApplications(ctx)
 
@@ -145,16 +144,14 @@ func (s *SparkSchedulerExtender) Predicate(ctx context.Context, args schedulerap
 		} else {
 			logger.Info("failed to schedule pod", svc1log.SafeParam("outcome", outcome), svc1log.SafeParam("reason", err.Error()))
 		}
-		s.wasteMetricsReporter.MarkFailedSchedulingAttempt(args.Pod, outcome)
-		return failWithMessage(ctx, args, err.Error())
+		return s.failWithMessage(ctx, outcome, args, err.Error())
 	}
 
 	if role == common.Driver {
 		appResources, err := sparkResources(ctx, args.Pod)
 		if err != nil {
 			logger.Error("internal error scheduling pod", svc1log.Stacktrace(err))
-			s.wasteMetricsReporter.MarkFailedSchedulingAttempt(args.Pod, failureInternal)
-			return failWithMessage(ctx, args, err.Error())
+			return s.failWithMessage(ctx, failureInternal, args, err.Error())
 		}
 		events.EmitApplicationScheduled(
 			ctx,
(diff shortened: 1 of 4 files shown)


internal/extender/resource.go
```diff
@@ -128,6 +132,7 @@ func (s *SparkSchedulerExtender) Predicate(ctx context.Context, args schedulerap
 	if err != nil {
 		msg := "failed to reconcile"
 		logger.Error(msg, svc1log.Stacktrace(err))
+		s.wasteMetricsReporter.MarkFailedSchedulingAttempt(args.Pod, failureInternal)
```
onursatici, line 135: we can move this into `failWithMessage`

internal/metrics/waste.go
```diff
@@ -49,94 +53,169 @@ var (
...
+		}
+	}
+}
+
+func (r *WasteMetricsReporter) onPodScheduled(pod *v1.Pod) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	info, ok := r.info[podKey{pod.Namespace, pod.Name}]
+	if !ok || info.demandCreationTime.IsZero() {
```
onursatici, line 200: lets return here early so we can de-indent the else block

internal/metrics/waste.go
```diff
@@ -49,94 +53,169 @@ var (
...
 		r.markAndSlowLog(pod, totalTimeNoDemand, time.Now().Sub(pod.CreationTimestamp.Time))
+	} else {
+		r.markAndSlowLog(pod, beforeDemandCreation, info.demandCreationTime.Sub(pod.CreationTimestamp.Time))
+
+		if !info.demandFulfilledTime.IsZero() {
```
onursatici, line 205: negate and return early

internal/metrics/waste.go
```diff
@@ -49,94 +53,169 @@ var (
...
+			if !info.lastFailedSchedulingAttemptInfo.attemptTime.IsZero() && info.lastFailedSchedulingAttemptInfo.attemptTime.After(info.demandFulfilledTime) {
```
onursatici, line 206: maybe extract `info.lastFailedSchedulingAttemptInfo` into a variable

internal/metrics/waste.go
```diff
@@ -31,16 +31,20 @@ import (
...
 )
 
 var (
 	beforeDemandCreation = tagInfo{
 		tag:              metrics.MustNewTag(schedulingWasteTypeTagName, "before-demand-creation"),
 		slowLogThreshold: 1 * time.Minute,
 	}
-	afterDemandFulfilled = tagInfo{
-		tag:              metrics.MustNewTag(schedulingWasteTypeTagName, "after-demand-fulfilled"),
+	afterDemandFulfilledNoFailures = tagInfo{
```
onursatici, line 42: I think we should keep `afterDemandFulfilled` and always mark it whenever we have a non zero demand fulfilled time, so we can compare

internal/metrics/waste.go
```diff
@@ -150,33 +229,66 @@ func (r *wasteMetricsReporter) markAndSlowLog(pod *v1.Pod, tag tagInfo, duration
...
 	defer r.lock.Unlock()
 	demand, ok := obj.(*v1alpha1.Demand)
 	if !ok {
 		svc1log.FromContext(r.ctx).Error("failed to parse obj as demand")
 		return
 	}
-	r.info[podKey{demand.Namespace, utils.PodName(demand)}] = demandInfo{
-		demandCreationTime: demand.CreationTimestamp.Time,
+	podKey := podKey{demand.Namespace, utils.PodName(demand)}
+	info, ok := r.info[podKey]
```
onursatici, line 261: it feels like to prevent all the branching on creating or merging the values in this map, we can change the type we store to `map[podkey]*schedulingMetricInfo`, I believe we can have a helper method to get a value from the map or create an empty struct if absent, and then do:
`r.info[podKey].demandCreationTime = demand.CreationTimestamp.Time`

internal/extender/resource.go
rkaram: Good point

PR #139, opened 2020-08-14
Node Overhead Calculation Refactor

rkaram:
Refactors the way we calculate node overhead to do the following:
* Make it dynamic in that it listens to pod informer events and updates requests on nodes accordingly rather than the current version which only recalculates the overhead every 30 seconds
* Only consider a pod as having returned its requested resources when it is deleted which is inline with the `kube-scheduler` calculation
* We also no longer fallback to the instance group median as we should now be able to accurately measure node overhead in realtime which makes the accurate calculation possible

The second point raises a potential issue: due to the fact that we were considering a pod as no longer requesting resources once it completes successfully or fails, we were undercounting overhead and were potentially optimistically scheduling pods on some nodes during binpacking. While more correct, the current solution means that in practice, the failure fit occurrences _might_ increase and affect scheduling times, which we should monitor.

### DIFF
--- a/internal/extender/overhead.go
+++ b/internal/extender/overhead.go
@@ -16,20 +16,18 @@ package extender
 
 import (
 	"context"
-	"sort"
 	"sync"
-	"time"
 
 	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
-	"github.com/palantir/k8s-spark-scheduler/internal/cache"
 	"github.com/palantir/k8s-spark-scheduler/internal/common"
-	werror "github.com/palantir/witchcraft-go-error"
+	"github.com/palantir/k8s-spark-scheduler/internal/common/utils"
 	"github.com/palantir/witchcraft-go-logging/wlog/svclog/svc1log"
-	"github.com/palantir/witchcraft-go-logging/wlog/wapp"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	coreinformers "k8s.io/client-go/informers/core/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
+	clientcache "k8s.io/client-go/tools/cache"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


internal/extender/overhead.go
```diff
@@ -39,151 +37,165 @@ var (
...
-		if p.Spec.SchedulerName != common.SparkSchedulerName {
-			// add all pods that this scheduler does not deal with to the non-schedulable overhead
-			o.addPodResourcesToGroupResources(ctx, rawNonSchedulableOverhead, p, instanceGroup)
+			if podRequestInfo.requests.CPU.AsDec().Cmp(oneCPU.AsDec()) > 0 || podRequestInfo.requests.Memory.AsDec().Cmp(oneGiB.AsDec()) > 0 {
+				svc1log.FromContext(ctx).Debug("Container with no resource reservation has high resource requests",
+					svc1log.SafeParam("podName", pod.Name),
+					svc1log.SafeParam("nodeName", pod.Spec.NodeName),
+					svc1log.SafeParam("CPU", podRequestInfo.requests.CPU),
+					svc1log.SafeParam("Memory", podRequestInfo.requests.Memory))
+			}
```
rkaram, line 155: I believe this will become too spammy if we start logging it here on every check to get overhead, so we should either remove it if it doesn't provide a lot of value, or consider logging it differently.

internal/extender/overhead.go
```diff
@@ -39,151 +37,165 @@ var (
...
-	podsWithRRs := make(map[string]bool, len(rrs))
-	for _, rr := range rrs {
-		for _, podName := range rr.Status.Pods {
-			podsWithRRs[podName] = true
-		}
+// computeNodeOverhead adds the requests of pods that don't have reservations and are counted as overhead.
+// This returns (overhead, nonSchedulableOverhead).
+func (o *OverheadComputer) computeNodeOverhead(ctx context.Context, nodeName string) (*resources.Resources, *resources.Resources) {
+	o.overheadLock.RLock()
+	defer o.overheadLock.RUnlock()
```
rkaram, line 127: Pushed this down here since we don't have pod movements across nodes, but if we think this might start conflicting too much with the informer updates and slow down scheduling, we can also lock once when getting overhead for all nodes.

internal/extender/overhead.go
```diff
@@ -39,151 +37,165 @@ var (
...
+	if _, ok := nodeRequests[pod.UID]; !ok {
+		return
+	}
+	delete(nodeRequests, pod.UID)
+	if len(nodeRequests) == 0 {
+		delete(o.resourceRequests, pod.Spec.NodeName)
+	}
 }
 
 func (o *OverheadComputer) addPodResourcesToGroupResources(ctx context.Context, groupResources map[string]resources.NodeGroupResources, pod *v1.Pod, instanceGroup string) {
```
onursatici, line 201: you can remove this method

internal/extender/overhead.go
onursatici: yea I think we can remove this log line

internal/extender/resourcereservations.go
```diff
@@ -80,6 +80,26 @@ func (rrm *ResourceReservationManager) GetResourceReservation(appID string, name
...
+// PodHasReservation returns if the passed pod has any reservation whether it is a resource reservation or a soft reservation
+func (rrm *ResourceReservationManager) PodHasReservation(ctx context.Context, pod *v1.Pod) bool {
+	appID, ok := pod.Labels[common.SparkAppIDLabel]
+	if !ok {
+		return false
+	}
+	if rr, ok := rrm.GetResourceReservation(appID, pod.Namespace); ok {
+		for _, rPodName := range rr.Status.Pods {
+			if pod.Name == rPodName {
+				return true
```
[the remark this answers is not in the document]
rkaram: I added a log line in the overhead calculation instead if the pod doesn't have a reservation, but the app does.

PR #174, opened 2021-07-12
Upgrade kube-scheduler from v1.13.1 to v1.15.12

chia7712:
the `v1.13.1` was release 2 years ago. It seems to me it is a bit out-of-date. Also, the max backoff in `v1.13.1` is 1 minute (https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/scheduler/util/backoff_utils.go#L103). That can raise scheduling latency when cluster gets enough resources (add more nodes or remove completed jobs). 

`v1.15.12` is the latest version which can accept `PodPriority=false`.

### DIFF
--- a/examples/extender.yml
+++ b/examples/extender.yml
@@ -94,7 +94,7 @@ spec:
       serviceAccountName: spark-scheduler
       containers:
       - name: kube-scheduler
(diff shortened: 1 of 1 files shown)


rkaram:
Thanks @chia7712

PR #179, opened 2021-09-14
[PR 3] Upgrade k8s lib versions to the same as KSSL

Chrisbattarbee:
This moves KSS from k8s 1.17.6 to 1.18.8, this aligns us with the latest version of KSSL so we can use it.

The minor version bump came with a few API breaks, so this PR also fixes those breaks.
Some functions have entirely disappeared, others have just changed signature.


internal/crd/utils.go
```diff
@@ -29,7 +30,7 @@ import (
 
 // CheckCRDExists checks if the given crd exists and is established
 func CheckCRDExists(crdName string, clientset apiextensionsclientset.Interface) (*apiextensionsv1beta1.CustomResourceDefinition, bool, error) {
-	crd, err := clientset.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crdName, metav1.GetOptions{})
+	crd, err := clientset.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.Background(), crdName, metav1.GetOptions{})
```
rkaram, line 33: pass in a context here rather than using `Background()`

rkaram (changes requested):
small nits but lgtm

rkaram (approved):

PR #195, opened 2021-11-29
Move off the hotpath

Chrisbattarbee:


rkaram (changes requested):
I don't think this approach will work because:
1. Many of the components in `initServer` do rely on the informers being synced while you've moved that to effectively happen at the end of this method. 
2. From looking at the code, at the end of `initServer`, I would expect the endpoints to work including the extender one, but this is not the case because all of it depends on an external go routine that implicitly syncs up informers when you set the clients. You could make this more explicitly with the Witchcraft initialisation pushing the information to a channel or similar.


What we're effectively trying to do in this case is say we have 2 servers that serve different functions but are put in the same one for practical reasons. There are 2 solutions that come to my mind:
1. Go all in on having 2 servers: create 2 Witchcraft servers, one for the webhook and one for the extender. I _think_ this will not easily work because they will both be on the same port and Witchcraft did not allow for customising that last I checked.
2. Separate the server initialisation into 2 big blocks: initialise the webhook and the extender server. The rough pseudocode I see is:
```
informers := createInformers()

initServer { 
    webhook := NewWebhook()
    extender := NewExtender(informers)
    addEndpoints(webhook, extender)
}
```

The `NewExtender` would have all the instantiation logic for the classes that require the informers to be instantiated and would return an error if they are not.

The problem with this is that the way we pass the signal changes and scheduler will no longer eagerly crash if it can't list objects on startup. We chould address those by having a Start() method on the `extender` which gets called as part of the startup but after witchcraft init (haven't checked feasibility in code)

Chrisbattarbee:
I agree with your comments. However I am unsure how the second approach you listed solves the problem, maybe I’m missing something. We have the following constraints:

1. Witchcraft starts serving all endpoints at the same time 
2. An endpoint which requires informers to be initialized to start correctly
3. Informers which require an endpoint to be serving traffic to initialize
4. One witchcraft server

There is no single threaded startup which satisfies these constraints 

I think if we want to be confident in our solution, we need two independent servers.

Chrisbattarbee:
Closing in favour of https://github.com/palantir/k8s-spark-scheduler/pull/196

PR #196, opened 2021-12-01
Move conversion webhook to separate container

Chrisbattarbee:
Previously the conversion webhook was hosted in the same witchcraft server as the predicate endpoint.

This lead to issues with initialisation as we have resource-reservation informers that need to sync before the predicate endpoint is usable. However, these informers rely on the conversion webhook being ready.

This PR solves this issue making the conversion webhook an entirely different service.


godel/config/dist-plugin.yml
```diff
@@ -1,4 +1,23 @@
...
+      os-archs:
+        - os: linux
+          arch: amd64
+    docker:
+      docker-builders:
+        webhook:
+          type: default
+          context-dir: resource-reservation-conversion-webhook/docker
+          tag-templates:
+            release: "{{Repository}}palantirtechnologies/resource-reservation-conversion-webhook:{{Version}}"
```
rkaram, line 17: Can we publish a new version only when we've changed that project? I don't think it's a big blocker for now, but I wonder if Godel has proper support for this, and if so, we should use it

internal/conversionwebhook/resource_reservation.go
```diff
@@ -41,25 +39,19 @@ func init() {
...
 	schedulerServicePort int32,
+	schedulerServiceContextPath string,
 ) (*v1.WebhookClientConfig, error) {
-	err := addConversionWebhookRoute(ctx, router)
-	if err != nil {
-		return nil, err
-	}
-
-	path := filepath.Join(server.ContextPath, webhookPath)
+	path := filepath.Join(schedulerServiceContextPath, webhookPath)
```
rkaram, line 51: If the services are now totally separate, why do we need to merge the paths? Can we do this at the service level?

rkaram (changes requested):
We have a conversion webhook running as part of the scheduler server today. That server is on a specific port. How's the migration plan? Are we going to expose this new webhook over a different service and start serving requests from it as we roll back the old nodes behind the old service?

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/golang/groupcache v0.0.0-20191227052852-215e87163ea7 // indirect
 	github.com/palantir/go-metrics v1.1.0
-	github.com/palantir/k8s-spark-scheduler-lib v0.2.20
+	github.com/palantir/k8s-spark-scheduler-lib v0.2.19
```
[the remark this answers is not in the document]
Chrisbattarbee: This version of KSSL contains a bug with two storage versions set to true

godel/config/dist-plugin.yml
Chrisbattarbee: Let me take a look!

Chrisbattarbee: From what I can see, it doesn't look like it supports this

internal/conversionwebhook/resource_reservation.go
Chrisbattarbee: We can have this all as config without joining. However, I'm not sure if we can set this at the service level, I think we'd need an ingress

Chrisbattarbee:
In short yep.

We can release a version of SSM that has the new webhook pods in a seperate deployment and a new service that points to them.

In this same release we can also point scheduler at this new service.

Therefore the existing scheduler pods will point at themselves like they previously did and new scheduler pods will point at the new webhook pods.

hack/dev/run-in-minikube.sh
```diff
@@ -16,6 +16,7 @@ $SCRIPT_ROOT/godelw docker build --verbose
 
 kubectl apply -f $SCRIPT_ROOT/examples/namespace.yml
 
-kubectl create configmap scheduler-secrets --namespace=spark --from-file="${SCRIPT_ROOT}/out/tmp/generated_keys/"
+kubectl create configmap scheduler-secrets --namespace=spark --from-file="${SCRIPT_ROOT}/out/generated_keys/"
```
rkaram, line 19: nit nit - maybe extract `${SCRIPT_ROOT}/out/generated_keys/` to a shared constant

rkaram (approved):

hack/dev/run-in-minikube.sh
Chrisbattarbee: 👍

PR #199, opened 2022-04-14
Add support for singleAZ scheduling and demand creation

Chrisbattarbee:
As a follow up to https://github.com/palantir/k8s-spark-scheduler-lib/pull/71 and https://github.com/palantir/k8s-spark-scheduler-lib/pull/70 this PR will allow scheduler to binpack pods into a single AZ and create demands that will be fulfilled in a single AZ.

The motivation being that applications scheduled in the same AZ will not incur cross AZ networking costs

## TODO
* [] Remove local replace once prior kss lib PR is merged

### DIFF
--- a/internal/extender/demand.go
+++ b/internal/extender/demand.go
@@ -69,17 +69,17 @@ func (s *SparkSchedulerExtender) createDemandForExecutor(ctx context.Context, ex
 			},
 		},
 	}
-	s.createDemand(ctx, executorPod, units)
+	s.createDemand(ctx, executorPod, units, false)
 }
 
-func (s *SparkSchedulerExtender) createDemandForApplication(ctx context.Context, driverPod *v1.Pod, applicationResources *sparkApplicationResources) {
+func (s *SparkSchedulerExtender) createDemandForApplication(ctx context.Context, driverPod *v1.Pod, applicationResources *sparkApplicationResources, enforceSingleZoneScheduling bool) {
 	if !s.demands.CRDExists() {
 		return
 	}
-	s.createDemand(ctx, driverPod, demandResources(applicationResources))
+	s.createDemand(ctx, driverPod, demandResources(applicationResources), enforceSingleZoneScheduling)
 }
 
-func (s *SparkSchedulerExtender) createDemand(ctx context.Context, pod *v1.Pod, demandUnits []demandapi.DemandUnit) {
+func (s *SparkSchedulerExtender) createDemand(ctx context.Context, pod *v1.Pod, demandUnits []demandapi.DemandUnit, enforceSingleZoneScheduling bool) {
 	instanceGroup, ok := internal.FindInstanceGroupFromPodSpec(pod.Spec, s.instanceGroupLabel)
 	if !ok {
 		svc1log.FromContext(ctx).Error("No instanceGroup label exists. Cannot map to InstanceGroup. Skipping demand object",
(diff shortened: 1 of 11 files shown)


internal/extender/demand.go
```diff
@@ -69,25 +69,25 @@ func (s *SparkSchedulerExtender) createDemandForExecutor(ctx context.Context, ex
 			},
 		},
 	}
-	s.createDemand(ctx, executorPod, units)
+	s.createDemand(ctx, executorPod, units, false)
```
VVKot, line 72: nit: that ties into what we discussed offline - you have no way of knowing what this `false` does apart from jumping to function definition. Doing Java-style constant like `DONT_ENFORCE_SINGLE_AZ_SCHEDULING` might alleviate that.

non-nit: is this truly what we want? The only codepath that leads to here is re-scheduling executors, but I don't know if that will always be true. At the very least, let's provide a justification.

Chrisbattarbee: On the nit: Yeah this is true, I ended up doing this way to avoid the proliferation of code paths. There are a few functions which wrap demand creation, we'd end up having duplication of those functions which ended I thought ended up being a little ugly. Happy to re-visit and see if I can reduce duplication here though


Non nit: Yep, we are explicitly not handling executor re-schedules in this iteration for a few reasons
* We do not currently have a way to create demands for a particular AZ http://github.com/palantir/k8s-spark-scheduler-lib/blob/02be893331748732b652e914a50bc20487161677/pkg/apis/scaler/v1alpha2/types_demand.go#L78-L91 
* If support for the above was added, it would need some more work in scheduler to associate the rescheduled executor with the single AZ application
* Metrics suggest that rescheduled executors are created fairly infrequent so we don't have much incentive to invest in this given the above

 
I'll add a comment so that it is clear!

Chrisbattarbee: Ah yes the constant makes a lot of sense here! Let me do that!

internal/metrics/metrics.go
```diff
@@ -220,8 +222,15 @@ func ReportCrossZoneMetric(ctx context.Context, driverNodeName string, executorN
...
+	crossAzTraffic := metrics.FromContext(ctx).Histogram(crossAzTraffic)
+	crossAzTraffic.Update(crossZonePairs)
+	totalTraffic := metrics.FromContext(ctx).Histogram(totalTraffic)
+	totalTraffic.Update(totalPairs)
+
+	// Needed because the mean is stripped from the metric logs of histograms by default, we need to explicitly update it
+	metrics.FromContext(ctx).GaugeFloat64(crossAzTrafficMean).Update(crossAzTraffic.Mean())
+	metrics.FromContext(ctx).GaugeFloat64(totalTrafficMean).Update(totalTraffic.Mean())
+
 	metrics.FromContext(ctx).Histogram(applicationZonesCount).Update(numberOfZones)
```
Chrisbattarbee, line 234: I'm really not a fan of this but it looks like there's not much we can do here, happy to discuss

rkaram: Haven't checked the `crossAzTraffic` calculation, but maybe we should discuss why the mean is what we care about in this instance

rkaram (changes requested):

internal/metrics/metrics.go
```diff
@@ -220,8 +222,19 @@ func ReportCrossZoneMetric(ctx context.Context, driverNodeName string, executorN
 	totalPairs := int64(totalNumPods * (totalNumPods - 1) / 2)
 	numberOfZones := int64(len(numPodsPerZone))
 
-	metrics.FromContext(ctx).Histogram(crossAzTraffic).Update(crossZonePairs)
-	metrics.FromContext(ctx).Histogram(totalTraffic).Update(totalPairs)
+	crossAzTraffic := metrics.FromContext(ctx).Histogram(crossAzTraffic)
```
rkaram, line 225: can you rename this so it doesn't conflict with the constant string?

rkaram: (same for `totalTraffic`)

rkaram (changes requested):

rkaram (approved):

PR #204, opened 2022-06-13
Use scratch as the base image for scheduler containers

Chrisbattarbee:
Reduce the size and attack surface of scheduler containers by using a base image of scratch

Tested locally on a minikube cluster

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,13 +1,25 @@
+# In the first stage, create our non-privileged user and directories
 FROM alpine:3.16
 
 ENV SERVICE_HOME=/opt/palantir/services/spark-scheduler
-ENV SPARK_SCHEDULER_STARTUP_CONFIG=$SERVICE_HOME/var/conf/install.yml
 
 RUN adduser -S -u 5001 spark-scheduler && \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,14 +1,26 @@
...
 
+FROM scratch
+
+ENV SERVICE_HOME=/opt/palantir/services/spark-scheduler
+
+# Copy the spark-scheduler user from the alpine stage
+COPY --from=0 /etc/passwd /etc/passwd
+
+# Copy the service home from the alpine stage
+COPY --from=0 $SERVICE_HOME $SERVICE_HOME
```
VVKot, line 20: nit: can't we just `ENV SERVICE_HOME=/opt/palantir/services/spark-scheduler` in this stage only? Same goes for SSCW image

VVKot (approved):
Looks good otherwise

docker/Dockerfile
VVKot: Ah nvm, we have to copy that because only `alpine` knows how to `mkdir`.

rbotarleanu (approved):
LGTM

PR #214, opened 2022-09-05
Schedule dynamically allocated executors in same AZ as the initial application

Chrisbattarbee:
This PR changes the default behaviour of scheduler to create a demand for additional executors requested by spark applications.

Additionally, if a single AZ scheduling algorithm is selected, scheduler will schedule executors in the same AZ as the existing application. In the event that there is no space in the AZ, it will create a demand which specifies that it must be fulfilled in the same AZ as the rest of the spark application


internal/extender/resource.go
```diff
@@ -47,7 +48,6 @@ const (
 	successRescheduled            = "success-rescheduled"
 	successAlreadyBound           = "success-already-bound"
 	successScheduledExtraExecutor = "success-scheduled-extra-executor"
-	failureFitExtraExecutor       = "failure-fit-extra-executor"
```
Chrisbattarbee, line 50: We always create demands for these pods now, so this is redundant

internal/common/utils/sets.go
```diff
@@ -51,7 +51,7 @@ func (s StringSet) Size() int {
 
 // ToSlice returns all the elements of StringSet as a slice of strings
 func (s StringSet) ToSlice() []string {
-	result := make([]string, s.Size())
+	result := make([]string, 0, s.Size())
```
Chrisbattarbee, line 54: This was a bug, which manifested when I tried to use this method to find unique AZs.

It would create a slice which looked like this if the set was <entry1, entry2> : ["", "", "entry1", "entry2"] as the second parameter to make is the number of initial elements, the third optional parameter is the capacity.

internal/extender/resource.go
```diff
@@ -415,6 +415,75 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
...
+
+		if zoneLabel == zone {
+			nodes = append(nodes, node)
+		}
+	}
+	svc1log.FromContext(ctx).Info("Filtered nodes in zone", svc1log.SafeParam("zone", zone), svc1log.SafeParam("nodes", nodes))
+	return nodes, nil
+}
+
+func (s *SparkSchedulerExtender) getCommonZoneForExecutorsApplication(ctx context.Context, executor *v1.Pod) (string, error) {
```
Chrisbattarbee, line 434: This function is too long. Break it down

internal/extender/resource.go
```diff
@@ -415,6 +415,75 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
...
+	}
+	svc1log.FromContext(ctx).Info("Found existing application pods", svc1log.SafeParam("applicationPodNames", applicationPodNames))
+	if err != nil {
+		return "", err
+	}
+
+	// Filter to scheduled pods
+	scheduledPods := make([]*v1.Pod, 0, len(applicationPods))
+	for _, pod := range applicationPods {
+		if pod.Status.Phase == v1.PodPending {
```
Chrisbattarbee, line 452: Change phase check to == `Running`. This ensures we dont try to check for completed pods

internal/extender/resource.go
```diff
@@ -437,15 +522,24 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
 			return name, successRescheduled, nil
 		}
 	}
-
-	if isExtraExecutor {
-		return "", failureFitExtraExecutor, werror.ErrorWithContextParams(ctx, "not enough capacity to schedule the extra executor")
+	zone, err = s.getZoneIfShouldScheduleInSingleAz(ctx, executor)
```
Chrisbattarbee, line 525: Should attempt to remove this call, duplicated from above

internal/extender/resource.go
```diff
@@ -426,6 +517,23 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
 	}
 	executorResources := &resources.Resources{CPU: sparkResources.executorResources.CPU, Memory: sparkResources.executorResources.Memory, NvidiaGPU: sparkResources.executorResources.NvidiaGPU}
 	availableNodes := s.getNodes(ctx, nodeNames)
+	zone, err := s.getZoneIfShouldScheduleInSingleAz(ctx, executor)
+	shouldScheduleInSingleAz := err == nil
```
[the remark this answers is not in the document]
VVKot: A better version (in my mind) would be:

```
shouldScheduleInSingleAz = false
if (doesBinpackingScheduleInSingleAz()) {
  log("Using single AZ scheduling")
  zone, err = getZone();
  if err != nil {
    return return "", failureInternal, err
  }
  nodes, err = filterNodesToZone(zone)
  if err != nil {
    return return "", failureInternal, err
  }
  shouldScheduleInSingleAz = true
} else {
  log("Not using single AZ scheduling")
}
```

VVKot (changes requested):
Overall idea makes sense but I feel that logic is tad convoluted to merge as-is

internal/extender/resource.go
```diff
@@ -415,6 +415,96 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
 	return availableNodes
 }
 
+func filterNodesToZone(ctx context.Context, initialNodes []*v1.Node, zone string) ([]*v1.Node, error) {
+	nodes := make([]*v1.Node, 0, len(initialNodes))
+	for _, node := range initialNodes {
+		zoneLabel, ok := node.Labels[v1.LabelTopologyZone]
+		if !ok {
+			return nil, werror.ErrorWithContextParams(ctx, "Could not read zone label from node, unable to make scheduling decisions based on AZ")
```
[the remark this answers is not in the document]
Chrisbattarbee: 👍 The entire way I view this is that if you are going to use a single AZ scheduling algo, then you need to have az information associated with your nodes.

internal/extender/resource.go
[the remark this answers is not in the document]
Chrisbattarbee: Done a refactoring here, will take another pass after we figure out issues with double reservations

internal/extender/resource.go
```diff
@@ -418,6 +422,101 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
...
+func (s *SparkSchedulerExtender) getSparkApplicationPodsForExecutor(ctx context.Context, executor *v1.Pod, executorSparkLabel string) ([]*v1.Pod, error) {
+	applicationPods, err := s.podLister.Pods(executor.Namespace).List(labels.Set(map[string]string{common.SparkAppIDLabel: executorSparkLabel}).AsSelector())
+	if err != nil {
+		return nil, err
+	}
+	applicationPodNames := make([]string, 0, len(applicationPods))
+	for _, pod := range applicationPods {
+		applicationPodNames = append(applicationPodNames, pod.Name)
+	}
+	svc1log.FromContext(ctx).Info("Found existing application pods", svc1log.SafeParam("applicationPodNames", applicationPodNames))
```
VVKot, line 516: nit: can we put this, `logRunningPods`, and other logging into a separate file? It makes functions longer and harder to read.

internal/extender/resource.go
```diff
@@ -418,6 +422,101 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
...
+func (s *SparkSchedulerExtender) getCommonZoneForExecutorsApplication(ctx context.Context, executor *v1.Pod) (string, bool, error) {
+	executorSparkLabel, ok := executor.Labels[common.SparkAppIDLabel]
+	if !ok {
+		return "", false, werror.ErrorWithContextParams(ctx, "Executor does not have a Spark app id label, could not create label selector")
+	}
+	applicationPods, err := s.getSparkApplicationPodsForExecutor(ctx, executor, executorSparkLabel)
+	if err != nil {
+		return "", false, err
+	}
+	scheduledPods := filterToRunningPods(ctx, applicationPods)
```
VVKot, line 455: Please comment why we're only looking at running pods.

internal/extender/resource.go
```diff
@@ -418,6 +422,101 @@ func (s *SparkSchedulerExtender) getNodes(ctx context.Context, nodeNames []strin
...
+	if azs.Size() > 1 {
+		return "", false, nil
+	}
+	if azs.Size() == 0 {
+		return "", false, werror.ErrorWithContextParams(ctx, "Application has no scheduled pods, can't make scheduling decisions based on AZ")
+	}
+	return azs.ToSlice()[0], true, nil
+}
+
+func filterToAZs(ctx context.Context, runningPods []*v1.Pod, s *SparkSchedulerExtender) (utils.StringSet, error) {
```
VVKot, line 470: nit: this seems to be `getPodsAZs`?

internal/extender/resource.go
```diff
@@ -429,6 +528,38 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
...
+		svc1log.FromContext(ctx).Info("Dynamic Allocation single AZ scheduling enabled, attempting to get zone to schedule into.")
+		zone, allPodsInSameAz, err := s.getCommonZoneForExecutorsApplication(ctx, executor)
+		if err != nil {
+			return "", "", err
+		}
+		if allPodsInSameAz {
+			svc1log.FromContext(ctx).Info("Only considering nodes from the zone",
+				svc1log.SafeParam("zone", zone))
+
+			availableNodes, err = filterNodesToZone(ctx, availableNodes, zone)
```
VVKot, line 544: I suppose we're assuming that if got here, `availableNodes` is going to be >0? Can we add a comment to clarify?

VVKot (changes requested):
Idea makes sense to me, let's add some tests/polish it slightly.

Would be great to add tests before cutting an RC to make sure we're not missing anything obvious.

internal/extender/resource.go
Chrisbattarbee: Yep, we'd throw in https://github.com/palantir/k8s-spark-scheduler/blob/1524603df98c80b53a6d71a1e534f399517fbacc/internal/extender/resource.go#L537

I'll add a comment

Chrisbattarbee:
We've already RC'd this and tested but I can add some tests

VVKot (approved):

PR #217, opened 2022-10-11
Bump k8s to v1.22.15

alibiyeslambek:


internal/extender/resource.go
```diff
@@ -275,8 +275,9 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
 			svc1log.SafeParam("nodeNames", nodeNames))
 		return driverReservedNode, success, nil
 	}
-	availableNodes, err := utils.ListWithPredicate(s.nodeLister, func(node *v1.Node) bool {
-		return helper.PodMatchesNodeSelectorAndAffinityTerms(driver, node)
```
alibiyeslambek, line 279: `PodMatchesNodeSelectorAndAffinityTerms` func is deprecated. Used the same approach as in k8s PR https://github.com/kubernetes/kubernetes/pull/99959/files

internal/common/utils/pods.go
```diff
@@ -115,7 +115,11 @@ func ListWithPredicate(nodeLister corelisters.NodeLister, predicate NodeConditio
 
 	var filtered []*v1.Node
 	for i := range nodes {
-		if predicate(nodes[i]) {
+		matches, err := predicate(nodes[i])
+		if err != nil {
+			return nil, err
+		}
+		if matches {
```
alibiyeslambek, line 122: This change is needed because we don't want to ignore errors caused by:
`PodMatchesNodeSelectorAndAffinityTerms ` -> `v1affinityhelper.GetRequiredNodeAffinity(driver).Match(node)`

rbotarleanu (approved):

PR #225, opened 2023-02-06
Add binpacking strategy to pack into the fewest bins possible

VVKot:
See https://github.com/palantir/k8s-spark-scheduler-lib/pull/98 for context


VVKot:
Closing to consolidate in https://github.com/palantir/k8s-spark-scheduler-lib/pull/100

PR #231, opened 2023-03-01
Make node sorting deterministic as it otherwise causes flakes

Alexis-D:
All things being equal, compare node names to break ties.

Example flake: https://app.circleci.com/pipelines/github/palantir/k8s-spark-scheduler/1857/workflows/017da816-ce66-4a03-898d-36ddfbe3d850/jobs/6103

Confirmed this works by running the test in a loop:
* with current logic it yields several failures
* with this PR it's all ✅


Chrisbattarbee (changes requested):
I think we should add the name to the schedule context instead of passing around a map to the comparison function. Especially because we do use this to schedule now in rare edge cases

Chrisbattarbee (approved):

PR #232, opened 2023-03-02
Update k8s scheduler lib and report packing metrics

pisarenko-net:
PR implements new metrics for packing efficiency. Packing is now run with 2 functions: the configured one and "pack tightly". The second round of packing is only used for reporting in metrics.

This depends on a new (unreleased) version of k8s-spark-scheduler-lib so this can't be merged until a new lib is released and dependencies are updated.

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -344,8 +344,8 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
 		return "", failureFit, werror.Error("application does not fit to the cluster")
 	}
 
-	metrics.ReportPackingEfficiency(ctx, s.binpacker.Name, *packingResultDefault)
-	metrics.ReportPackingEfficiency(ctx, "tightlyPack", *packingResultTight)
+	metrics.ReportPackingEfficiency(ctx, s.binpacker.Name, availableNodesSchedulingMetadata, packingResultDefault)
+	metrics.ReportPackingEfficiency(ctx, "tightlyPack", availableNodesSchedulingMetadata, packingResultTight)
 
 	s.removeDemandIfExists(ctx, driver)
 	metrics.ReportCrossZoneMetric(ctx, driverNode, executorNodes, availableNodes)
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -332,7 +332,6 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
 		svc1log.SafeParam("candidateExecutorNodes", executorNodeNames),
 		svc1log.SafeParam("driverNode", packingResult.DriverNode),
 		svc1log.SafeParam("executorNodes", packingResult.ExecutorNodes),
-		svc1log.SafeParam("packingEfficiency", packingResult.ExecutorNodes),
 		svc1log.SafeParam("binpacker", s.binpacker.Name))
 	if !packingResult.HasCapacity {
 		s.createDemandForApplicationInAnyZone(ctx, driver, applicationResources)
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -448,10 +448,12 @@ func (s *SparkSchedulerExtender) rescheduleExecutor(ctx context.Context, executo
 	}
 	executorResources := &resources.Resources{CPU: sparkResources.executorResources.CPU, Memory: sparkResources.executorResources.Memory, NvidiaGPU: sparkResources.executorResources.NvidiaGPU}
 	availableNodes := s.getNodes(ctx, nodeNames)
-	usages := s.resourceReservationManager.GetReservedResources()
-	usages.Add(s.overheadComputer.GetOverhead(ctx, availableNodes))
-	availableResources := resources.AvailableForNodes(availableNodes, usages)
-	availableNodesSchedulingMetadata := resources.NodeSchedulingMetadataForNodes(availableNodes, usages)
+
+	usage := s.resourceReservationManager.GetReservedResources()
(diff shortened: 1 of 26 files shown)


pisarenko-net:
Disclaimer: please review the metrics interactions carefully. I am not sure if anything else is required (like an explicit emit, or rate limiting). If there's a good primer on metrics please send me a link.

internal/extender/resource.go
```diff
@@ -304,14 +306,27 @@ func (s *SparkSchedulerExtender) selectDriverNode(ctx context.Context, driver *v
...
+		ctx,
+		applicationResources.driverResources,
+		applicationResources.executorResources,
+		applicationResources.minExecutorCount,
+		driverNodeNames,
+		executorNodeNames,
+		availableNodesSchedulingMetadata)
+	driverNode, executorNodes, hasCapacity := packingResultDefault.DriverNode, packingResultDefault.ExecutorNodes, packingResultDefault.HasCapacity
+	// as an experiment, binpack with tight pack (report in metrics but don't use this result)
+	packingResultTight := SelectBinpacker("tightlyPack").BinpackFunc(
```
Chrisbattarbee, line 321: I think we can do this but lets keep the PR tightly scoped and do this in another PR

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+const (
+	packingEfficiencyMetricName = "foundry.spark.scheduler.packing_efficiency"
+
+	packingResourceTagKey = "foundry.spark.scheduler.packing_resource"
+	cpuTagValue           = "CPU"
+	memoryTagValue        = "Memory"
+	gpuTagValue           = "GPU"
+	maxTagValue           = "Max"
+
+	packingEfficiencyFunctionNameTagKey = "foundry.spark.scheduler.packing_function"
```
Chrisbattarbee, line 57: To be consistent we should not have `_`

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+func reportableEfficiencyFromAvgPackingEfficiency(efficiency *binpack.AvgPackingEfficiency) reportableEfficiency {
+	return reportableEfficiency{
+		CPU:    efficiency.CPU,
+		Memory: efficiency.Memory,
+		GPU:    efficiency.GPU,
+	}
+}
+
+const (
+	packingEfficiencyMetricName = "foundry.spark.scheduler.packing_efficiency"
```
Chrisbattarbee, line 49: To be consistent we should not have `_`

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+// ReportPackingEfficiency report packing efficiency metrics for a single packing result.
+func ReportPackingEfficiency(
+	ctx context.Context,
+	packingFunctionName string,
+	nodesSchedulingMetadata resources.NodeGroupSchedulingMetadata,
+	packingResult *binpack.PackingResult) {
+
+	contextTags := metrics.TagsFromContext(ctx)
```
Chrisbattarbee, line 78: I think its unlikely that we'll want all of these tags, the cardinality of the metric will be massive

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+	registry := metrics.NewRootMetricsRegistry()
```
Chrisbattarbee, line 79: We dont want to create a metrics registry here

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+
+	packingFunctionTag := metrics.MustNewTag(packingEfficiencyFunctionNameTagKey, packingFunctionName)
+
+	// report packing efficiencies per node
+	for _, efficiency := range packingResult.PackingEfficiencies {
+		hostTag := HostTag(ctx, efficiency.NodeName)
```
Chrisbattarbee, line 85: Tagging by host is going to have extremely high cardinality

internal/metrics/binpack.go
```diff
@@ -0,0 +1,117 @@
...
+		emitMetrics(registry, contextTags, hostTag, packingFunctionTag, reportableEfficiencyFromPackingEfficiency(efficiency))
+	}
+
+	// report avg packing efficiency for all nodes at once
+	efficiency := computeAvgPackingEfficiencyForResult(nodesSchedulingMetadata, packingResult)
+	avgHostTag := HostTag(ctx, avgEfficiencyNodeName)
+	emitMetrics(registry, contextTags, avgHostTag, packingFunctionTag, reportableEfficiencyFromAvgPackingEfficiency(&efficiency))
```
Chrisbattarbee, line 92: Im a bit confused as to why we're reporting twice here?

Chrisbattarbee (changes requested):

internal/metrics/binpack.go
pisarenko-net: This is the pattern we've used in scaler telemetry: report efficiency per node and then separately for all nodes at once. The upside of reporting by node is that we can then see what the efficiency in every node. Obviously, comes with an increase in amount of data. We could start with just the avg efficiency, disregarding node stats.

pisarenko-net:
Thank you for the review Chris, I'm going to fix merge conflicts and incorporate feedback.

internal/extender/resource.go
pisarenko-net: Ack!

internal/metrics/binpack.go
pisarenko-net: Do I understand correctly that the important tags, such as instance-group, deployment are populated automatically when calling `metrics.FromContext(ctx)...`?

Chrisbattarbee (approved):

PR #248, opened 2023-03-24
Add driver/executor collocation metric

Alexis-D:
==COMMIT_MSG==
Add driver/executor collocation metric
==COMMIT_MSG==

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -209,9 +209,9 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
 	}
 }
 
-// ReportDriverCollocationMetrics reports metrics around whether driver collocation with executors. This is meant to be used
+// ReportDriverCollocationMetric reports metrics around whether driver collocation with executors. This is meant to be used
 // during the initial scheduling.
-func ReportDriverCollocationMetrics(ctx context.Context, instanceGroup string, driverNodeName string, executorNodeNames []string) {
+func ReportDriverCollocationMetric(ctx context.Context, instanceGroup string, driverNodeName string, executorNodeNames []string) {
 	instanceGroupTag := InstanceGroupTag(ctx, instanceGroup)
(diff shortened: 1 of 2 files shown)


internal/metrics/metrics.go
```diff
@@ -56,12 +56,14 @@ const (
 	podInformerDelay                          = "foundry.spark.scheduler.informer.delay"
 	schedulingWaste                           = "foundry.spark.scheduler.scheduling.waste"
 	schedulingWastePerInstanceGroup           = "foundry.spark.scheduler.scheduling.wasteperinstancegroup"
+	driverCollocation                         = "foundry.spark.scheduler.scheduling.drivercollocation"
```
VVKot, line 59: nit: `driverexecutorinitialcollocation`? I want to convey that's the state when Spark application starts, even though it might change later.

internal/metrics/metrics.go
```diff
@@ -203,6 +209,31 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
 	}
 }
 
+// ReportDriverCollocationMetric reports metrics around whether driver collocation with executors. This is meant to be used
```
VVKot, line 212: nit nit: ...whether driver is collocated...

internal/metrics/metrics.go
```diff
@@ -203,6 +209,31 @@ func (s *ScheduleTimer) Mark(ctx context.Context, role, outcome string) {
...
+// during the initial scheduling.
+func ReportDriverCollocationMetric(ctx context.Context, instanceGroup string, driverNodeName string, executorNodeNames []string) {
```
VVKot, line 214: Same as above, `ReportInitialDriverCollocationMetric`?

internal/metrics/metrics.go
Alexis-D: yup, shouldn't blindly trust goland refactor

VVKot (approved):

PR #253, opened 2023-04-07
improvement: Refactor server initialization to expose a plugin point to end to end tests

k-simons:
- Refactor server initialization to expose a plugin point to end to end tests 
- In the future we can use `InitServerWithClients` to pass through K8s clients in memory to ensure that the full server initialization works and all informers and objects jobs will be ran

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/cmd/clients.go
+++ b/cmd/clients.go
@@ -1,0 +1,76 @@
+// Copyright (c) 2019 Palantir Technologies. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Chrisbattarbee (approved):
This looks good to me, thanks for putting this up.

Good that we're actually throwing here now: https://github.com/palantir/k8s-spark-scheduler/pull/253/files#diff-ee2560d627cd77f1563972217a33ac39a4bf56fbf30d7f560fbb4199b9f70197R124

PR #254, opened 2023-04-07
improvement: Remove resource reservation cache and instead just use a store backed by a lister

k-simons:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


internal/reservations/store.go
```diff
@@ -0,0 +1,94 @@
...
+	return d.resourceReservationLister.List(labels.Everything())
+}
+
+func (d *defaultStore) Create(ctx context.Context, obj *v1beta2.ResourceReservation) error {
+	result, err := d.sparkschedulerV1beta2Interface.ResourceReservations(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	d.allowListerToSync(ctx, result)
+	return err
```
Alexis-D, line 66:
```suggestion
	return nil
```

internal/reservations/store.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (d *defaultStore) Update(ctx context.Context, obj *v1beta2.ResourceReservation) error {
+	result, err := d.sparkschedulerV1beta2Interface.ResourceReservations(obj.GetNamespace()).Update(ctx, obj, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	d.allowListerToSync(ctx, result)
+	return err
```
Alexis-D, line 75:
```suggestion
	return nil
```

internal/extender/resourcereservations.go
```diff
@@ -77,8 +78,12 @@ func NewResourceReservationManager(
 }
 
 // GetResourceReservation returns the resource reservation for the passed pod, if any.
-func (rrm *ResourceReservationManager) GetResourceReservation(appID string, namespace string) (*v1beta2.ResourceReservation, bool) {
-	return rrm.resourceReservations.Get(namespace, appID)
+func (rrm *ResourceReservationManager) GetResourceReservation(ctx context.Context, appID string, namespace string) (*v1beta2.ResourceReservation, bool) {
+	r, err := rrm.resourceReservationStore.Get(ctx, namespace, appID)
+	if err != nil {
```
Alexis-D, line 83:
```suggestion
	return r, err == nil
```

Alexis-D (commented):
Overall seems like a good idea (and would allow us to revert the #262 hack), but unless I'm missing something won't this lead to a situation where we create an rr, we silently timeout trying to see that update, then later on we fail to see that rr because we're lagging behind?

(i guess you could argue that we already have this issue when running multiple spark scheduler nodes anyway?)

PR #255, opened 2023-04-07
improvement: Move the demand processing behind a single interface that can be shared

k-simons:
- Move the demand processing behind a single interface that can be shared 
- Right now demand manipulation are a combination of static methods and using the raw `SafeDemandCache` struct which makes it impossible to test or mock
- This moves this operations behind a single interface
- Move out `binpacker` to its own package (`binpacker`) so it can be shared
- Move the common object types into their own package (`types`) so it can be shared

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/internal/extender/resource.go
+++ b/internal/extender/resource.go
@@ -16,6 +16,8 @@ package extender
 
 import (
 	"context"
+	binpacker2 "github.com/palantir/k8s-spark-scheduler/internal/binpacker"
+	"github.com/palantir/k8s-spark-scheduler/internal/demands"
 	"time"
 
 	demandapi "github.com/palantir/k8s-spark-scheduler-lib/pkg/apis/scaler/v1alpha2"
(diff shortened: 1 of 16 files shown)


internal/extender/resource.go
```diff
@@ -24,9 +24,11 @@ import (
 	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
 	"github.com/palantir/k8s-spark-scheduler/config"
 	"github.com/palantir/k8s-spark-scheduler/internal"
+	binpacker2 "github.com/palantir/k8s-spark-scheduler/internal/binpacker"
```
VVKot, line 27: 👀

internal/extender/unschedulablepods.go
```diff
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/palantir/k8s-spark-scheduler-lib/pkg/resources"
+	binpacker2 "github.com/palantir/k8s-spark-scheduler/internal/binpacker"
```
VVKot, line 22: 👀

VVKot (approved):
Let's rename `binpacker2` but otherwise good to merge!

internal/extender/unschedulablepods.go
k-simons: Fixed

PR #256, opened 2023-04-11
Add time to bind metrics

Alexis-D:
## Before this PR

#250 helps quantify how many resources are taken by unbound reservations, however it doesn't help us understand how long reservation stay unbound.

## After this PR

==COMMIT_MSG==
Add time to bind metrics
==COMMIT_MSG==

## Possible downsides?

This doesn't capture reservations that never get bound, but that shouldn't really happen under normal circumstances.

### DIFF
--- a/internal/extender/resourcereservations.go
+++ b/internal/extender/resourcereservations.go
@@ -340,10 +340,10 @@ func (rrm *ResourceReservationManager) bindExecutorToResourceReservation(ctx con
 	// only report the metric the first time the reservation is bound
 	if _, ok := resourceReservation.Status.Pods[reservationName]; !ok {
 		// this is the k8s server time, so the duration we're computing only makes sense if clocks are reasonably kept in sync
-		creationMicros := resourceReservation.CreationTimestamp.Time.UnixMicro()
-		now := time.Now().UnixMicro()
-		durationMicros := now - creationMicros
-		metrics.ReportTimeToBindMetrics(ctx, durationMicros)
+		creationMicros := resourceReservation.CreationTimestamp.Time
+		now := time.Now()
+		duration := now.Sub(creationMicros)
+		metrics.ReportTimeToBindMetrics(ctx, duration)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/metrics/metrics.go
```diff
@@ -364,3 +367,11 @@ func (dct *SoftReservationCompactionTimer) MarkCompactionComplete(ctx context.Co
 func IncrementSingleAzDynamicAllocationPackFailure(ctx context.Context, zone string) {
 	metrics.FromContext(ctx).Counter(singleAzDynamicAllocationPackFailureCount, ZoneTag(ctx, zone)).Inc(1)
 }
+
+// ReportTimeToBindMetrics reports how long it takes between a reservation being created and pods being bound to said reservation.
+func ReportTimeToBindMetrics(ctx context.Context, duration time.Duration) {
+	timeToBindHist := metrics.FromContext(ctx).Histogram(timeToBind)
+	timeToBindHist.Update(duration.Microseconds())
```
VVKot, line 374: nit: we typically use nanoseconds

internal/extender/resourcereservations.go
```diff
@@ -335,6 +336,14 @@ func (rrm *ResourceReservationManager) bindExecutorToResourceReservation(ctx con
 	if err != nil {
 		return werror.WrapWithContextParams(ctx, err, "failed to update resource reservationName", werror.SafeParam("reservationName", reservationName))
 	}
+
+	// only report the metric the first time the reservation is bound
+	if _, ok := resourceReservation.Status.Pods[reservationName]; !ok {
```
VVKot, line 341: I have convinced myself this is the check we want. We don't care why a pod was bound to a reservation, and yet we're still in this method - it would be quite complex to handle all cases. Let's start with first bound time, but please update the metric to name to indicate that.

https://github.com/palantir/k8s-spark-scheduler/blob/2112d580b54a922c5ca5fd7cc4723c83a4272201/internal/extender/resourcereservations.go#L361-L382

VVKot (approved):

PR #265, opened 2023-04-13
Log details about reservation that remain unbound for too long

Alexis-D:


internal/extender/resourcereservations.go
```diff
@@ -367,6 +371,16 @@ func (rrm *defaultResourceReservationManager) bindExecutorToResourceReservation(
 		if !creationTime.IsZero() {
 			duration := time.Since(creationTime)
 			metrics.ReportTimeToFirstBindMetrics(ctx, duration)
+
+			if duration > slowLogDuration {
+				svc1log.FromContext(ctx).Warn(
+					"Resource reservation was unbound for a while",
```
VVKot, line 377: nit: "Time to first executor bind to resource reservation is above threshold"

internal/extender/resourcereservations.go
```diff
@@ -367,6 +371,16 @@ func (rrm *defaultResourceReservationManager) bindExecutorToResourceReservation(
...
+					svc1log.SafeParam("duration", duration.String()),
```
VVKot, line 378: I don't think you need `.String()`.

internal/extender/resourcereservations.go
```diff
@@ -367,6 +371,16 @@ func (rrm *defaultResourceReservationManager) bindExecutorToResourceReservation(
...
+					svc1log.SafeParam("appID", resourceReservation.Labels[v1beta1.AppIDLabel]),
+					svc1log.SafeParam("node", node),
+					svc1log.SafeParam("executor", executor.Name),
```
VVKot, line 381: Let's also log reservation name.

VVKot: And executor namespace.

VVKot (approved):

